PR #10, opened 2024-11-19
Fix issue where annotations are not properly traversed for Python 3.8

rasmysamy:
This PR properly traverses Python 3.8 annotations to recursively apply `_adjust_annotations`, which fixes bugs where Pydantic models are not properly serialized.


ptomecek (approved):
Approved, but note that the aim is to completely remove python 38 support soon as it is EOL. Also worth noting that all of the annotation adjustment stuff should go away once this issue is closed: https://github.com/pydantic/pydantic/issues/9670.

PR #12, opened 2024-11-19
Fix bug where GraphEvaluator would not return the result of the model being directly called

rasmysamy:
Also fixes a typo in CompositePublisher


ccflow/publishers/composite.py
```diff
@@ -37,7 +37,7 @@ class CompositePublisher(BasePublisher, Generic[PydanticModelType]):
     )
 
     models_as_dict: bool = Field(True, description="Whether to expand fields that contain pydantic models into dictionaries.")
-    options: PydanticDictOptions = Field(PydanticDictOptions(), dscription="Options for iterating through the pydantic model.")
+    options: PydanticDictOptions = Field(PydanticDictOptions(), description="Options for iterating through the pydantic model.")
```
timkpaine, line 40: good catch!

Issue #26, opened 2025-02-01
Add Polars result type


ptomecek:
**Is your feature request related to a problem? Please describe.**
Would like to be able to natively represent polars dataframes (eager or lazy) as a result type from a `CallableModel`

**Describe the solution you'd like**
Either a single result type for both types of frames, or two different result types (one for eager one for lazy). Optionally, we should be able to specify a schema that will get validated (as is done for the Arrow result type). 

**Describe alternatives you've considered**
Purely using arrow tables works well for eager frames, but there are use cases where it is convenient to have callable models construct lazy frames instead to defer execution.

ptomecek:
Closed by #35

PR #33, opened 2025-02-25
Pydantic validation of Narwhals types

ptomecek:
This will simplify #26, while also leading to support of several other underlying libraries without explicit work required!

Since this code was written natively against pydantic v2, the approach is much cleaner than what we use in these places:
 - https://github.com/Point72/ccflow/blob/main/ccflow/exttypes/arrow.py
 - https://github.com/Point72/ccflow/blob/main/ccflow/exttypes/pandas.py

With this functionality, I think we can deprecate some of those types.

### DIFF
--- a/ccflow/exttypes/narwhals.py
+++ b/ccflow/exttypes/narwhals.py
@@ -1,0 +1,225 @@
+import difflib
+import inspect
+import pprint
+from collections import OrderedDict
+from dataclasses import dataclass
+from typing import Annotated, Any, Optional, Union, get_args, get_origin
+
+import narwhals.stable.v1 as nw
+from pydantic import AfterValidator, ConfigDict, GetCoreSchemaHandler, TypeAdapter
+from pydantic_core import core_schema
+
+__all__ = (
+    "FrameValidator",
+    "DataFrameValidator",
+    "LazyFrameValidator",
+    "FrameT",
+    "DataFrameT",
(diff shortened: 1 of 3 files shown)


timkpaine (approved):
We might consider adding it to (or just removing) the `full` optional dependencies section

ccflow/exttypes/narwhals.py
```diff
@@ -0,0 +1,225 @@
+import difflib
+import inspect
+import pprint
+from collections import OrderedDict
+from dataclasses import dataclass
+from typing import Annotated, Any, Optional, Union, get_args, get_origin
+
+import narwhals.stable.v1 as nw
```
hintse, line 8: Since narwhals is an "optional" dependency at the moment, should we raise a more meaningful error if we get an import error here (i.e. ask users in install it manually)?

ptomecek: Narwhals is lightweight, so I will just add it to the dependencies section.

timkpaine: we conditionally export in `exttypes/__init__.py` https://github.com/Point72/ccflow/blob/58bb2d3f6ae650f0a51f36ee2d07d8e7136f3416/ccflow/exttypes/__init__.py#L7

timkpaine (approved):

PR #35, opened 2025-02-25
Add narwhals result types

ptomecek:
Addresses #26 in a more general way (with support for other backends).

### DIFF
--- a/ccflow/result/narwhals.py
+++ b/ccflow/result/narwhals.py
@@ -1,0 +1,28 @@
+import narwhals.stable.v1 as nw
+from pydantic import Field
+
+from ..base import ResultBase
+from ..exttypes.narwhals import DataFrameT, FrameT
+
+__all__ = (
+    "NarwhalsFrameResult",
+    "NarwhalsDataFrameResult",
+)
+
+
+class NarwhalsFrameResult(ResultBase):
+    df: FrameT = Field(description="Narwhals DataFrame or LazyFrame")
+
+    def collect(self) -> "NarwhalsDataFrameResult":
+        """Collects the result into a NarwhalsDataFrameResult."""
+        if isinstance(self.df, nw.LazyFrame):
+            return NarwhalsDataFrameResult(df=self.df.collect(), **self.model_dump(exclude={"df", "type_"}))
+        return NarwhalsDataFrameResult(df=self.df, **self.model_dump(exclude={"df", "type_"}))
+
+
+class NarwhalsDataFrameResult(NarwhalsFrameResult):
+    df: DataFrameT = Field(description="Narwhals eager Dataframe")
+
+    def collect(self) -> "NarwhalsDataFrameResult":
+        """Collects the result into a NarwhalsDataFrameResult."""
+        return self
--- a/ccflow/tests/result/test_narwhals.py
+++ b/ccflow/tests/result/test_narwhals.py
@@ -1,0 +1,71 @@
+from typing import Annotated
+
+import narwhals.stable.v1 as nw
+import polars as pl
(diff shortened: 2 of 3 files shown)


timkpaine (dismissed):

ccflow/result/narwhals.py
```diff
@@ -0,0 +1,24 @@
...
+__all__ = (
+    "NarwhalsResult",
+    "NarwhalsDataFrameResult",
+)
+
+
+class NarwhalsResult(ResultBase):
+    df: FrameT = Field(description="Narwhals DataFrame or LazyFrame")
+
+    def collect(self) -> "NarwhalsDataFrameResult":
```
NeejWeej, line 16: NarwhalsDataFrameResult IS a NarwhalsResult, so if we call "collect" on a NarwahslDataFrameResult, why do we make a new NarwhalsDataFrameResult instead of returning self?

NeejWeej:
Maybe useful to add conversions to the other Result Types (like ArrowResult here: https://github.com/Point72/ccflow/blob/main/ccflow/result/pyarrow.py#L14 )

so that like, we can validate generic CallableModel annotations switching between the types (though maybe Im misunderstanding), like here:  
https://github.com/Point72/ccflow/blob/main/ccflow/callable.py#L539

Though those could also be separate PR's (if we do decide to go that way)

ccflow/result/narwhals.py
```diff
@@ -0,0 +1,24 @@
...
+        """Collects the result into a NarwhalsEagerResult."""
+        if isinstance(self.df, nw.LazyFrame):
+            return NarwhalsDataFrameResult(df=self.df.collect(), **self.model_dump(exclude={"df", "type_"}))
+        return NarwhalsDataFrameResult(df=self.df, **self.model_dump(exclude={"df", "type_"}))
+
+
+class NarwhalsDataFrameResult(NarwhalsResult):
```
feussy, line 23: I'm a bit on the fence about the naming convention, our lazy result is a `NarwhalsResult` and our collected DF is a `NarwhalsDataFrameResult`. I almost feel like something like a `NarwhalsLazyResult` and `NarwhalsResult` for the lazy and collected versions respectively may make more sense?

ptomecek: `NarwhalsResult` holds either a lazy or an eager frame - it corresponds to the `nw.FrameT` type. The `NarwhalsDataFrameResult` corresponds to eager-only - it corresponds to the `nw.DataFrameT` type. There is no pure narwhals type for lazy frame only. However, perhaps we should do `NarwhalsResult->NarwhalsFrameResult` to make the naming clearer?

ptomecek:
> Maybe useful to add conversions to the other Result Types (like ArrowResult here: https://github.com/Point72/ccflow/blob/main/ccflow/result/pyarrow.py#L14 )
> 
> so that like, we can validate generic CallableModel annotations switching between the types (though maybe Im misunderstanding how this works), like here: https://github.com/Point72/ccflow/blob/main/ccflow/callable.py#L539
> 
> Though those could also be separate PR's (if we do decide to go that way)

I will do some more cleanup of the `ArrowResult` separately, but I want the narwhals results not to refer to the arrow results as I now think of them as a bit deprecated.

ccflow/result/narwhals.py
feussy: Understood - then I think just renaming `NarwhalsResult` as you suggested may make sense

ptomecek: This has been renamed to `NarwhalsFrameResult`

feussy (approved):

PR #50, opened 2025-03-07
Fix #49. Make sure to use right context when raising unresolved model exception during registry load.

### LINKED ISSUE #49: Registry lookup errors during config loading can sometimes be suppressed by accident
**Describe the bug**

If a sub-registry accidentally references an object from a parent registry with a local path (instead of an absolute path or using `../`, then it will fail to resolve the path. However, it won't raise an exception - it will just drop it silently from the registry

**Expected behavior**
It should raise an exception (preferably the original exception, with a traceback), but definitely not fail silently. 

What is happening is that it fails to resolve under the registry lookup context here: https://github.com/Point72/ccflow/blob/main/ccflow/base.py#L605
However, when we try to reproduce the exception (to get the full traceback), we are not running it under the same context: https://github.com/Point72/ccflow/blob/main/ccflow/base.py#L631
This means that the evaluation which is supposed to generate the exception actually succeeds, and no exception is raised.

ptomecek:


ccflow/tests/test_base_registry.py
```diff
@@ -496,6 +500,48 @@ def test_model_lookup_error(self):
...
+                        "_target_": "ccflow.tests.test_base_registry.MyNestedModel",
+                        "x": "foo",  # Relative path, should be ../foo or /foo, so should raise
+                        "y": {"a": "test2", "b": 2.0},
+                    },
+                },
+            }
+        )
+        r = ModelRegistry(name="test")
+        msg = "RegistryKeyError('Could not find model foo in RootModelRegistry')\n"
+        "RegistryKeyError('Could not find model foo in RootModelRegistry')\n"
```
NeejWeej, line 522: would it be possible to add in the error like, where it was looking for foo (like in subregistry/bar)

ptomecek: hydra provides this for us in the InstantiationException, just not explicitly tested for.

NeejWeej (approved):

PR #66, opened 2025-05-09
Add log_result option to LoggingEvaluator with formatting options

ptomecek:


ccflow/evaluators/common.py
```diff
@@ -92,13 +95,36 @@ def make_result():
...
+    This is used by the LoggingEvaluator to control how the result is formatted.
+    """
+
+    arrow_as_polars: bool = Field(
+        False,
+        description="Whether to convert pyarrow tables to polars tables for formatting, as arrow formatting does not work well with large tables or provide control over options",
+    )
+    pformat_config: Dict[str, Any] = Field(_DEFAULT_PFORMAT_CONFIG, description="pformat config to use for formatting data")
+    polars_config: Dict[str, Any] = Field({}, description="Polars config to use for formatting polars frames")
+    pandas_config: Dict[str, Any] = Field({}, description="pandas config to use for formatting pandas objects")
```
hintse, line 116: Why does "Polars" (line above) have upper case P but "pandas" has lower case p?

ccflow/evaluators/common.py
```diff
@@ -92,13 +95,36 @@ def make_result():
...
 class LoggingEvaluator(EvaluatorBase):
     """Evaluator that logs information about evaluating the callable.
 
     It logs start and end times, the model name, and the context."""
 
     log_level: int = Field(logging.DEBUG, description="The log level for start/end of evaluation")
-    verbose: bool = Field(True, description="Whether to output the model defintion as part of logging")
+    verbose: bool = Field(True, description="Whether to output the model definition as part of logging")
+    log_result: bool = Field(False, description="Whether to log the result of the evaluation")
+    format_config: FormatConfig = Field(FormatConfig(), description="Configuration for formatting the result of the evaluation if log_result=True")
```
hintse, line 127: Just out of curiosity, is there a difference between `default=FormatConfig()` and `default_factory=FormatConfig`?

hintse (dismissed):

ccflow/evaluators/common.py
ptomecek: Fixed (all lowercase)

ccflow/evaluators/common.py
ptomecek: you only need default_factory if you need to delay the creation of the default (i.e. if it depends on datetime.now). Pydantic  makes a copy of the default value so unlike in standard python, you don't need to worry about mutable default values.

PR #71, opened 2025-05-28
Make pickle of BaseModel consistent between processes

ptomecek:
Workaround for https://github.com/pydantic/pydantic/issues/11603 for `ccflow.BaseModel`.


NeejWeej:
Does this work with Annotations or fields where we use `Field` from pydantic?

ptomecek:
> Does this work with Annotations or fields where we use `Field` from pydantic?

Yes, I improved the test case here to illustrate.

PR #93, opened 2025-07-29
Add NarwhalsFilePublisher

hintse:


ccflow/publishers/file.py
```diff
@@ -1,13 +1,15 @@
...
 
+import narwhals.stable.v1 as nw
 import pandas as pd
 import yaml
 from cloudpathlib import AnyPath
 from pydantic import Field, field_validator
 from typing_extensions import Literal, override
 
 from ..exttypes import JinjaTemplate
+from ..exttypes.narwhals import FrameT
```
ptomecek, line 12: If you only want eager frames here, you want `DataFrameT`. If you use `FrameT`, you'll have to call `collect()` if a lazy frame is passed in.

PR #120, opened 2025-09-30
Better exception handling for registry loading

ptomecek:
This PR fixes a couple of problems with error reporting when loading configs into the registry.

1. The previous logic reported only the first exception. Sometimes this was a resolution error because the referenced model failed to load, and the user wants to see *that* exception so they know what to fix. However, there could be a huge amount of resolution errors (and they could be chained together). 
  We fix this by splitting the exceptions into resolution and non-resolution exceptions. 
    * If there are non-resolution exceptions, we raise all of these on python 3.11+ using `ExceptionGroup` (on <3.11 we just raise the first one). In this case, we skip any resolution exceptions. 
    * If there are only resolution exceptions, we just raise all of these (or the first one on <3.11) as there might be some circular logic. 

2. A common mistake is that users use a relative path, i.e. `foo`, to refer to another model from within a subregistry, when they mean to use an absolute path, i.e. `/foo`, to look up an object from the global registry. Previously this would return an exception. Now we detect when `/foo` actually exists and suggest to the user that maybe they meant to do that instead. If there is no `/foo`, there is no change in the message.


ccflow/base.py
```diff
@@ -597,7 +598,7 @@ class _ModelRegistryLoader:
     def __init__(self, overwrite: bool):
         self._overwrite = overwrite
 
-    def _make_subregistries(self, cfg, registries: List[ModelRegistry]) -> List[Tuple[List[ModelRegistry], str, DictConfig]]:
+    def _make_subregistries(self, cfg, registries: List[ModelRegistry]) -> List[Tuple[List[ModelRegistry], str, DictConfig, Optional[Exception]]]:
```
hintse, line 601: Where is the new item in the tuple (`Optional[Exception]`) used?

hintse: Ah, so it's only for the type consistency of `models_to_register` in `load_config()`.

PR #125, opened 2025-09-30
Use pydantic serialize_as_any flag instead of adjusting annotations

ptomecek:
This is now possible due to https://github.com/pydantic/pydantic/issues/9049
Still need to sort out some test failures.


ptomecek:
Replaced by #215

PR #127, opened 2025-10-01
Utilize polymorphic_serialization instead of Metaclass to adjust annotations

NeejWeej:
Utilize the `polymorphic_serialization` flag introduced in this PR:

https://github.com/pydantic/pydantic/pull/12518

This PR will fail to build until a pydantic release has this feature in it.


ccflow/callable.py
```diff
@@ -426,6 +437,36 @@ def __call__(self) -> ResultType:
...
+    def _serialize_model_evaluation_context(self, handler: SerializerFunctionWrapHandler):
+        try:
+            return handler(self)
+        except ValueError as exc:
+            msg = str(exc)
+            if "Circular reference" not in msg and "id repeated" not in msg:
+                raise
+            # Minimal, stable representation sufficient for cache-key tokenization
+            try:
+                model_repr = self.model.model_dump(mode="python", serialize_as_any=True, by_alias=True)
```
ptomecek, line 456: Not sure this will handle `mode="json"` passed upstread, or `model_dump_json`.

NeejWeej: I added a test file to test for these cases.

pyproject.toml
```diff
@@ -44,7 +44,7 @@ dependencies = [
     "orjson",
     "pandas",
     "pyarrow",
-    "pydantic>=2.6,<3",
+    "pydantic>=2.35,<3",
```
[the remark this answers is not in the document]
NeejWeej: Ah yup, I think that was the pydantic-core version where this came from.

PR #128, opened 2025-10-02
Add mixin to allow pulling key-value pairs from python objects

NeejWeej:
Add `PathKeyResolverMixin` which allows specifying some key/value pairs from python code via a `PyObjectPath`. This allows offloading some of the configuration to python, getting benefits from:
easier testability, cohesiveness of related configurations, and deduplication. 

This functionality is opt-in, since base `ccflow.BaseModel` classes do not inherit from it via default.

### DIFF
--- a/ccflow/utils/mixin.py
+++ b/ccflow/utils/mixin.py
@@ -1,0 +1,53 @@
+from typing import ClassVar, List, Optional
+
+from pydantic import PrivateAttr, model_validator
+
+from .path_resolve_spec import PathResolveSpec
+
+__all__ = ("PathKeyResolverMixin",)
+
+
+class PathKeyResolverMixin:
+    """Resolve configuration values from a python-import path with optional nested key traversal.
+
+    Supported fields (either top-level or within a `ccflow` dict):
+      - `ccflow_path`: import path string (coerced to PyObjectPath)
+      - `ccflow_keys`: list[str|int] or dotted string for nested traversal (supports list indices)
+      - `ccflow_merge`: one of {"resolved_wins", "explicit_wins", "raise_on_conflict"}
+      - `ccflow_filter_extras`: bool; default True; drop resolved keys not present on the model
+
+    Optional: subclasses may set `ccflow_allowed_prefixes` to restrict import paths.
+      - `ccflow_allowed_prefixes` is a ClassVar[List[str]] of allowed import-path prefixes. If set, any
+        `ccflow_path` not starting with one of these prefixes raises a ValueError during validation.
+        This provides a lightweight allowlist to constrain dynamic imports in configs.
+    """
+
+    ccflow_allowed_prefixes: ClassVar[Optional[List[str]]] = None
+    __ccflow_source__: dict = PrivateAttr(default_factory=dict)
+
+    @model_validator(mode="wrap")
+    def _resolve_and_attach(cls, v, handler):
+        # Only operate on dict inputs; otherwise construct directly
+        values = v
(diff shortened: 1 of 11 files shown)


ccflow/utils/mixin.py
```diff
@@ -0,0 +1,53 @@
...
+            node = spec.traverse(obj)
+            mapping = spec.as_mapping(node)
+            mapping = spec.filter_extras_map(mapping, getattr(cls, "model_fields", {}))
+            values = spec.merge_into(values, mapping)
+
+        # Construct instance
+        result = handler(values)
+
+        # Attach debug metadata if spec was provided
+        if isinstance(result, object) and spec is not None:
```
NeejWeej, line 48: maybe there should be an option to disable this?

ccflow/utils/path_resolve_spec.py
```diff
@@ -0,0 +1,160 @@
...
+        if self.merge == "resolved_wins":
+            values.update(mapping)
+            return values
+        elif self.merge == "explicit_wins":
+            for k, v in mapping.items():
+                if k not in values:
+                    values[k] = v
+            return values
+        else:
+            conflicts = [k for k, v in mapping.items() if k in values and values[k] != v]
```
NeejWeej, line 146: arguably we could do this check in the in every case, and log the results.

NeejWeej:
> This is a relatively big change, and I have to think about other ways of accomplishing this.
> 
> For example, instead of the configs referencing python paths of other configs, if those other configs were in the registry already, then the existing mechanisms would work to reference them. Then the question becomes how to take some model instances that are defined in python and add them as named objects to the registry. I think this may be doable leveraging hydra's `_call_` abilities (i.e. instead of `_target_`) - i.e. you call a function in python that returns you some object and stick that in the registry under a given name. Then other registry objects can reference it as usual, and the logic for instantiating those objects just lives in python functions...

Good point. I added a function `from_python` that can be called with `_target_` and just pulls an object from python (and since it's in hydra, we get it added to the registry for free)

ccflow/base.py
```diff
@@ -324,17 +323,13 @@ def _is_config_subregistry(value):
 
 
 def model_alias(model_name: str) -> BaseModel:
```
ptomecek: Why not just import model_alias from "compose" into this file?

ccflow/compose.py
```diff
@@ -0,0 +1,101 @@
...
+    if indexer:
+        for key in indexer:
+            obj = obj[key]
+    return obj
+
+
+def update_from_base(
+    base: Any = None,
+    *,
+    target_class: Optional[Any] = None,
```
ptomecek: Should we have a type hint here? `Union[str, Type, None]`

ptomecek (changes requested):

ccflow/compose.py
```diff
@@ -0,0 +1,101 @@
...
+        indexer: ["a", "b"]
+    """
+    obj = PyObjectPathTA.validate_python(py_object_path).object
+    if indexer:
+        for key in indexer:
+            obj = obj[key]
+    return obj
+
+
+def update_from_base(
```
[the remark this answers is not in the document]
NeejWeej: I kept it "base" to be intentionally generic. It doesnt have to be a model, but can also be a dict for "base"

I'd be down with `update_from_template`

ccflow/base.py
NeejWeej: Fixed, just used `TYPE_CHECKING` to defer the import

ptomecek (approved):

PR #130, opened 2025-10-06
Update PolarsExpression to just be a type annotation on pl.Expr

NeejWeej:
without subclassing.

This follows the approach highlighted here: https://docs.pydantic.dev/latest/concepts/types/#handling-third-party-types

on the pydantic documentation.

This will be a breaking change because `PolarsExpression` no longer has a `validate` classmethod

### DIFF
--- a/ccflow/__init__.py
+++ b/ccflow/__init__.py
@@ -1,4 +1,4 @@
-__version__ = "0.6.9"
+__version__ = "0.6.10"
 
 from .arrow import *
 from .base import *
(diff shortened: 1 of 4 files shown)


ccflow/exttypes/polars.py
```diff
@@ -7,12 +7,12 @@
 import polars as pl
 from packaging import version
 from pydantic import TypeAdapter
-from typing_extensions import Self
+from typing_extensions import Annotated, Self
```
timkpaine, line 10: typing

PR #139, opened 2025-10-29
Ensure CallableModelGenericType works when impl is separated from generic

timkpaine:


ccflow/tests/test_callable.py
```diff
@@ -339,6 +339,32 @@ def __call__(self, context):
...
+            pass
+
+        class MyCallable(MyCallableImpl):
+            @Flow.call
+            def __call__(self, context: NullContext) -> GenericResult[int]:
+                return GenericResult[int](value=42)
+
+        class DateRangeImplAnalyticDirect(MyCallableBase[NullContext, GenericResult[int]]):
+            @Flow.call
+            def __call__(self, context: NullContext) -> GenericResult[int]:
```
ptomecek: Can we add a unit test to check that it raises if the Context/Result from the generic don't match the context/result type hints in the `__call__` method

PR #141, opened 2025-10-30
Allow partial generics in inheritance tree

timkpaine:


ccflow/callable.py
```diff
@@ -574,55 +574,49 @@ def _validate_callable_model_generic_type(cls, m, handler, info):
...
+                    raise TypeError(
+                        f"Return type annotation {orig_return_typ} on __call__ does not match result_type {new_result_type} defined by CallableModelGenericType"
+                    )
+
+                # Set on class
+                cls._result_type = new_result_type
+
+        subtypes = cls.__pydantic_generic_metadata__["args"]
+        if subtypes:
+            TypeAdapter(Type[subtypes[0]]).validate_python(m.context_type)
```
ptomecek: Instantiating new type adapters dynamically is pretty slow. We should cache these somewhere (I think we have a pattern elsewhere, or in csp, that does this).

PR #146, opened 2025-11-01
Bugfix for wrap validator invocation during assignment

timkpaine:


timkpaine:
https://github.com/Point72/ccflow/pull/153

PR #170, opened 2025-12-31
Add ability to auto-generate context classes from function signature

NeejWeej:


ptomecek:
From an implementation perspective, I get why it looks like this:
```python
            @Flow.call
            @dynamic_context
            def __call__(self, *, x: int, y: str = "default") -> GenericResult:
                return GenericResult(value=f"{x}-{y}")
```
But I feel like from a user perspective, it would be nice to have it look like this
```python
            @Flow.call(dynamic_context=True)
            def __call__(self, *, x: int, y: str = "default") -> GenericResult:
                return GenericResult(value=f"{x}-{y}")
```

timkpaine:
why does the user have to indicate that its a dynamic context at all manually? isn't this implied by not having a single arg `context` of type inheriting from BaseContext?

PR #195, opened 2026-04-14
feat: native tokenization engine for BaseModel

ptomecek:
## Summary

Replace dask-based tokenization with a standalone, composable engine that works with any pydantic BaseModel.

### API

- `model.model_token` — deterministic hex digest of a model's data (and optionally behavior)
- `__ccflow_tokenizer__` ClassVar — swap the tokenizer engine per-class or globally
- `__ccflow_tokenizer_deps__` ClassVar — declare standalone function dependencies for behavior hashing

### Architecture

Two independent axes of composition:

- **FunctionCollector** (which functions to hash) × **SourceTokenizer** (how to hash them)
- `DefaultTokenizer()` — data only
- `DefaultTokenizer.with_bytecode()` — data + bytecode-hashed own methods (default for behavior)
- `DefaultTokenizer.with_ast()` — data + AST-normalized own methods

### What changed

| File | Change |
|------|--------|
| `ccflow/utils/tokenize.py` | Core engine — SourceTokenizer/FunctionCollector ABCs, normalize_token singledispatch, DefaultTokenizer |
| `ccflow/base.py` | `model_token` property, `__ccflow_tokenizer__` ClassVar, cache invalidation |
| `ccflow/callable.py` | `ModelEvaluationContext.model_token` — strips transparent evaluator layers |
| `ccflow/evaluators/common.py` | Simplified `cache_key()` to use `model_token` directly |
| `ccflow/utils/__init__.py` | Export new types |
| `pyproject.toml` | Remove dask from tokenization path |
| `docs/wiki/Tokenization.md` | Comprehensive wiki page |
| `ccflow/tests/utils/test_tokenize.py` | 182 test cases |

### Extension points

1. `normalize_token.register(MyType)` — custom type handlers via singledispatch
2. `__ccflow_tokenize__()` — custom canonical form for any object
3. `FunctionCollector` / `SourceTokenizer` subclasses — custom strategies
4. `Field(default_factory=...)` — inject external state (package versions, file checksums, env vars) into tokens

### Limitations

- No transitive dependency tracking (only own methods + explicit deps)
- Bytecode tokens change across Python minor versions
- AST tokens change on variable renames
- See wiki page for full false-hit/false-miss tables

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -35,7 +35,7 @@ jobs:
           - '3.11'
         dependencies:
           - ''
-          - '"pandas<2" "numpy<2" "xarray<2025.09.0" "dask<2024.7.0"'
+          - '"pandas<2" "numpy<2" "xarray<2025.09.0"'
           - '"pandas<3"'
           - '"pandas<4"'
 
--- a/ccflow/base.py
+++ b/ccflow/base.py
@@ -31,6 +31,7 @@ from typing_extensions import Self
 
 from .exttypes.pyobjectpath import PyObjectPath
 from .local_persistence import register_ccflow_import_path, sync_to_module
+from .utils.tokenize import DefaultTokenizer, Tokenizer, normalize_token
 
 log = logging.getLogger(__name__)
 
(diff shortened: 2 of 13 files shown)


NeejWeej:
Maybe we could narrow `model_token` down.
 Something like:
  - model_token means data token only
  - behavior becomes separate: behavior_version() or model_behavior_token
  - cache key combines them
  That removes a lot of conceptual confusion.
  So for a callable model:
  model.model_token         # explicit field-data snapshot
  model.behavior_version()  # code semantics
  Then the cache-key generator composes both.

ptomecek:
> Maybe we could narrow `model_token` down. Something like:
> 
> * model_token means data token only
> * behavior becomes separate: behavior_version() or model_behavior_token
> * cache key combines them
>   That removes a lot of conceptual confusion.
>   So for a callable model:
>   model.model_token         # explicit field-data snapshot
>   model.behavior_version()  # code semantics
>   Then the cache-key generator composes both.

Part of me wonders if we should even both with caching/having it on the model. We could remove any access from the model itself, and just have everything go through the existing tokenizer classes. It would then be the responsibility of the evaluator to do any caching. 

The drawback of that is that I think it's then more difficult for people to debug issues because it's harder to reproduce exactly what was used (i.e. from a notebook)...

PR #198, opened 2026-04-23
fix: Orphaned sub-registry name resolution in `get_registered_names()`

feussy:
## Summary

Fixes `get_registered_names()` returning `[]` for models whose parent sub-registry has been orphaned by an `add(..., overwrite=True)` call.

When a sub-registry is replaced in the root (e.g. during a config reload), `add()` removes `(root, name)` from the old sub-registry's `_registrations`, breaking the chain. Models still referencing the old sub-registry dead-end at the empty `_registrations` and can no longer resolve their names — silently breaking any downstream logic that depends on registry-path lookup (e.g. cache evaluators keyed by model name).

### `_RegistryMixin._was_registered`
- New `PrivateAttr(default=False)` added alongside `_registrations`
- Set to `True` in `ModelRegistry.add()` the first time a model is added to any parent
- Distinguishes an *orphaned* sub-registry (was once registered, then replaced) from a *fresh/local* sub-registry (created but never added to any parent) — the latter must remain invisible to name resolution, as existing semantics require

### `get_registered_names()` — orphan fallback branch
- New `elif isinstance(self, ModelRegistry) and self._was_registered and self.name:` branch before `return []`
- Orphaned sub-registries return `[f"/{self.name}"]`, reconstructing their path from the immutable `.name` attribute
- Works at arbitrary nesting depth automatically: only the topmost orphaned registry hits the new branch; intermediate registries still have intact `_registrations` pointing to it, so the existing recursion assembles the full path correctly
- Three guards prevent false positives: `isinstance` (never fires for plain `BaseModel` overwrites), `_was_registered` (never fires for local unrooted registries), `self.name` (belt-and-suspenders against root, whose `name` is `""`)

## What's NOT included
- No changes to `ModelRegistry.add()` de-registration logic — the removal of old registrations on overwrite is intentional and tested
- No changes to callers of `get_registered_names()` — all existing callers handle a non-empty return correctly

## Tests
1 new test `test_orphaned_subregistry_resolves_name` covering the 2-level deep case (`/foo/bar/model_name`), with registry cleanup in `finally`. All 40 existing registry tests pass unchanged — including `test_add_twice` and `test_add_two_places` which verify that local unrooted sub-registries remain invisible.

`test_pickle_consistency` in `test_base_serialize.py` was updated to reflect that `_was_registered` is now included in `__pydantic_private__` during pickling. The hard-coded expected bytes were regenerated accordingly; the test's purpose (proving deterministic pickle output across runs) is unchanged.

### DIFF
--- a/ccflow/base.py
+++ b/ccflow/base.py
@@ -64,17 +64,34 @@ class _RegistryMixin:
         """Return the set of registrations that has happened for this model"""
         return self._registrations.copy()
 
-    def get_registered_names(self) -> List[str]:
-        """Return the set of names for this model in the root registry."""
+    def get_registered_names(self, include_orphaned: bool = False) -> List[str]:
+        """Return the set of names for this model in the root registry.
+
+        Args:
+            include_orphaned: If True, also return names for models whose parent
+                sub-registry has been orphaned (replaced via add(..., overwrite=True)).
+                An orphaned registry is no longer reachable from root but retains its
+                original .name. Default is False, which preserves legacy behavior:
+                models in orphaned sub-registries return [].
+
+        Note: when include_orphaned=True, the returned path is reconstructed from the
+        orphaned registry's .name and is not guaranteed to resolve via a live root lookup.
+        The invariant that (reg, name) in model._registrations implies
+        reg._models[name] is model does not hold for orphaned registries.
+        """
         if self._registrations:
             full_names = []
             for registry, name in self._registrations:
-                registry_names = registry.get_registered_names()
+                registry_names = registry.get_registered_names(include_orphaned=include_orphaned)
                 for registry_name in registry_names:
                     full_names.append(REGISTRY_SEPARATOR.join([registry_name, name]))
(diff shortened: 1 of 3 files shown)


ptomecek:
I think we need a flag to differentiate the different questions the method might be ansering
- Is this model currently registered in the root registry, and if so, under what names?
- Was this model ever registered in the root registry and if so, under what names?

I don't know what I would call such a parameter, `current` perhaps or `include_previous`? I would make the default equal to the past behavior.

feussy:
> I think we need a flag to differentiate the different questions the method might be ansering
> 
> * Is this model currently registered in the root registry, and if so, under what names?
> * Was this model ever registered in the root registry and if so, under what names?
> 
> I don't know what I would call such a parameter, `current` perhaps or `include_previous`? I would make the default equal to the past behavior.

Agreed, have added a flag and associated test that preserves existing behavior by default

ptomecek (approved):

PR #208, opened 2026-05-12
Native tokenization engine, drop dask runtime dep

ptomecek:
Replaces the `dask.base.normalize_token` / `tokenize` re-exports with a native singledispatch engine, and drops `dask` from runtime dependencies (adds `cloudpickle`).

Builds on #196 — `compute_data_token`, `compute_cache_token`, and `compute_behavior_token` keep their existing semantics; the underlying data-tokenization backend changes from dask to the native engine.

## What's covered

Structural handlers for: stdlib primitives, `datetime` / `date` / `time` / `timedelta`, `Decimal`, `UUID`, `pathlib`, `Enum`, `complex`, `Ellipsis`, `slice`, `functools.partial`, `MappingProxyType`, `OrderedDict` (order-preserving), `types.MethodType` / `MethodWrapperType`, `types.CodeType`, `types.FunctionType` (defaults / kwdefaults / closure cells participate), `type`, and `pydantic.BaseModel`. Lazy-registered handlers for `numpy.ndarray` (object-dtype arrays recurse element-wise) and `numpy.generic`. Unknown types fall back to `cloudpickle`, raising `TypeError` if pickling fails.

`tokenize()` keeps its variadic `(*args, **kwargs)` signature for drop-in compatibility with the dask API.

Cycle detection via a module-level `ContextVar` returns a stable `(\"__cycle__\", type_name)` marker instead of raising `RecursionError`.

## Performance vs dask

Rough one-off comparisons (μs/op):

| Workload | dask | native | ratio |
|---|---:|---:|---:|
| Small primitive dict | 8.9 | 5.8 | **0.65x** |
| Pydantic BaseModel (6 fields) | 1910 | 8.7 | **0.005x (~220x faster)** |
| Nested BaseModel (100 sub-models) | 3778 | 704 | **0.19x (~5x faster)** |
| Function with comprehension | 38.6 | 7.8 | **0.20x (~5x faster)** |
| ndarray int64, 1k | 5.6 | 9.1 | 1.6x |
| ndarray float64, 1M | 441 | 4539 | **10x slower** |
| ndarray object, 1k strings | 13.4 | 236 | **18x slower** |
| pd.Series int64, 1k | 11.2 | 31.6 | 3x slower |

**Wins** are concentrated where ccflow actually computes cache keys: pydantic models and their compositions (~220x), small dicts, and functions. **Losses** are on bulk numpy / pandas data — dask streams a contiguous view directly into the hasher and has fast paths for string-dtype arrays; this implementation does \`tobytes() → repr() → encode() → sha256\` and recurses element-wise for object arrays. Worth a follow-up if multi-MB arrays become typical cache inputs.

## Tests

818 passing (+128 new), including:
- determinism / structural-portability regression tests for closures, comprehensions, nested defs, lambdas, object-dtype ndarrays
- collision tests for `OrderedDict` order-sensitivity, distinct method bindings
- cycle-detection coverage (self-referential dicts, lists, pydantic models, shared sub-objects not false-flagged as cycles)
- variadic `tokenize` signature semantics

## Out of scope

- `BaseModel.model_token` property, frozen-model Merkle caching, `__ccflow_tokenize__` extension hook — slated for a follow-up PR
- Streaming hash for large numpy arrays — perf follow-up

### DIFF
--- a/ccflow/tests/utils/test_tokenize.py
+++ b/ccflow/tests/utils/test_tokenize.py
@@ -1,16 +1,25 @@
 """Tests for tokenize helpers used by cache keys."""
 
+import enum as _enum
+import re
+from collections import OrderedDict
+from datetime import date, datetime, time, timedelta
+from decimal import Decimal
+from functools import partial
+from pathlib import Path, PurePosixPath
+from types import MappingProxyType
+from uuid import UUID
+
+import numpy as np
+import pandas as pd
 import pytest
+from pydantic import BaseModel as _PlainPydantic
 
 from ccflow.callable import CallableModel, ContextBase, EvaluatorBase, ModelEvaluationContext
 from ccflow.context import NullContext
 from ccflow.evaluators.common import cache_key
 from ccflow.result import GenericResult
-from ccflow.utils.tokenize import compute_behavior_token, compute_cache_token, compute_data_token
-
-# ---------------------------------------------------------------------------
-# Data token
-# ---------------------------------------------------------------------------
+from ccflow.utils.tokenize import compute_behavior_token, compute_cache_token, compute_data_token, normalize_token, tokenize
 
 
 class TestComputeDataToken:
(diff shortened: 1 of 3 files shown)


ccflow/tests/utils/test_tokenize.py
```diff
@@ -628,3 +641,650 @@ def g(self):
         assert t_base != t_sub
         # But base's cached token is unaffected
         assert compute_behavior_token(Base) == t_base
+
+
+# ---------------------------------------------------------------------------
```
timkpaine: remove

ptomecek: done, thanks for flagging. The only one left is the one that separates data hashing from behavior hashing logic in that file

PR #211, opened 2026-05-15
Handle pickling for generic pydantic models, fixes #210

### LINKED ISSUE #210: [BUG] `GenericResult[T]` Cloudpickle Fails In Fresh Processes When Serialized By Global Reference
## Description

`GenericResult[int]` is a runtime-created Pydantic generic specialization. When
that specialization is created at module top level, Pydantic registers it in
`ccflow.result.generic` under the generated name `GenericResult[int]` so pickle
can serialize it by global reference.

That global registration is process-local. If one process cloudpickles a
`GenericResult[int]` instance by reference, and a fresh process or Ray worker
tries to unpickle it before that worker has created the same specialization,
unpickle fails with:

```text
AttributeError: Can't get attribute 'GenericResult[int]' on
<module 'ccflow.result.generic' ...>
```

It is a base serialization issue for
runtime-created Pydantic generic result classes. `@Flow.model` (from #206) and Ray make the
bug much easier to hit because generated model state or distributed worker
payloads can contain `GenericResult[T]` instances without importing/re-evaluating
the source module in a way that materializes the specialization first.

The behavior is sensitive to where `GenericResult[int]` is first evaluated:

- If `GenericResult[int]` is evaluated at module top level, Pydantic installs a
  module global and cloudpickle may serialize by global reference. This fails in
  a fresh process that lacks that process-local global.
- If `GenericResult[int]` is created only inside a function, cloudpickle may
  serialize the dynamic class by value instead. That can pass, which masks the
  issue and makes the failure look import-order dependent.

### Repro 1: plain cloudpickle, no Ray, no `@Flow.model`

1. Save or run this script:

```python
# /// script
# requires-python = ">=3.12"
# dependencies = [
#   "ccflow",
#   "cloudpickle",
# ]
# ///

import base64
import subprocess
import sys


def run_case(label, dumps_kwargs):
    creator_script = f"""
import base64
import cloudpickle
from ccflow import GenericResult

print(base64.b64encode(cloudpickle.dumps(GenericResult[int](value=5), **{dumps_kwargs!r})).decode())
"""
    creator = subprocess.run([sys.executable, "-c", creator_script], capture_output=True, text=True)
    blob = creator.stdout.strip()

    loader_script = f"""
import base64
import cloudpickle
import ccflow.result.generic as generic_module
from ccflow import GenericResult

print("has GenericResult[int] before unpickle:", hasattr(generic_module, "GenericResult[int]"))
obj = cloudpickle.loads(base64.b64decode({blob!r}))
print("unpickled:", obj)
"""
    result = subprocess.run([sys.executable, "-c", loader_script], capture_output=True, text=True)
    print(f"\\n=== {label} ===")
    print("returncode:", result.returncode)
    print("stdout:", result.stdout.strip() or "<empty>")
    print("stderr:", result.stderr.strip() or "<empty>")


run_case("cloudpickle default protocol", {})
run_case("cloudpickle protocol=5", {"protocol": 5})
```

2. Run it with uv:

```bash
uv run python repro_generic_result_cloudpickle.py
```

3. Observe that both default cloudpickle protocol and `protocol=5` fail in the
   fresh loader process.

### Repro 2: Ray worker, no `@Flow.model`

1. Save or run this script:

```python
# /// script
# requires-python = ">=3.12"
# dependencies = [
#   "ccflow",
#   "cloudpickle",
#   "ray",
# ]
# ///

import base64
from typing import Any

import cloudpickle
import ray

from ccflow import GenericResult


# Toggle this flag:
# - True: materializes GenericResult[int] at module top level and reproduces the failure.
# - False: avoids the top-level global-reference path and can pass, which shows why
#   this bug is sensitive to where the specialization is first evaluated.
top_level_eval = True

if top_level_eval:
    SPECIALIZED_RESULT = GenericResult[int]


@ray.remote
def load_payload(label: str, encoded_payload: str) -> tuple[str, bool, str]:
    import base64
    import traceback

    import ccflow.result.generic as generic_module
    import cloudpickle
    from ccflow import GenericResult  # noqa: F401

    has_specialization = hasattr(generic_module, "GenericResult[int]")
    try:
        obj = cloudpickle.loads(base64.b64decode(encoded_payload))
    except Exception:
        return label, has_specialization, traceback.format_exc()
    return label, has_specialization, f"unpickled: {obj!r}"


def make_payload(*, dumps_kwargs: dict[str, Any]) -> str:
    if top_level_eval:
        payload = cloudpickle.dumps(SPECIALIZED_RESULT(value=5), **dumps_kwargs)
    else:
        payload = cloudpickle.dumps(GenericResult[int](value=5), **dumps_kwargs)
    return base64.b64encode(payload).decode()


ray.init(num_cpus=1)
try:
    cases = {
        "cloudpickle default protocol": make_payload(dumps_kwargs={}),
        "cloudpickle protocol=5": make_payload(dumps_kwargs={"protocol": 5}),
    }
    refs = [load_payload.remote(label, payload) for label, payload in cases.items()]
    for label, has_specialization, result in ray.get(refs):
        print(f"\\n=== {label} ===")
        print("worker has GenericResult[int] before unpickle:", has_specialization)
        print(result.strip())
finally:
    ray.shutdown()
```

2. In this repo, run Ray repros with uv targeting the active venv so Ray workers
   inherit `ray` correctly:

```bash
source .venv/bin/activate
uv run --active python repro_generic_result_ray_cloudpickle.py
```

3. With `top_level_eval = True`, observe the worker-side unpickle failure.

4. Set `top_level_eval = False` and run again. The script can pass because the
   specialization is no longer serialized through the same top-level
   global-reference path. This confirms the bug is tied to Pydantic's
   process-local global registration side effect, not to `@Flow.model`.

## Expected Behavior

A `GenericResult[T]` instance should be portable through cloudpickle across a
fresh process or Ray worker.

In particular, this should work regardless of whether `GenericResult[int]` has
already been materialized in the receiving process:

```python
payload = cloudpickle.dumps(GenericResult[int](value=5))
restored = cloudpickle.loads(payload)
assert restored == GenericResult[int](value=5)
```

The same should hold when the payload is sent to a Ray worker.

## Actual Behavior

The receiving process imports `GenericResult`, but `GenericResult[int]` is not
present in `ccflow.result.generic` before unpickle:

```text
has GenericResult[int] before unpickle: False
```

Plain cloudpickle/fresh-process output:

```text
=== cloudpickle default protocol ===
returncode: 1
stdout: has GenericResult[int] before unpickle: False
stderr: Traceback (most recent call last):
  File "<string>", line 11, in <module>
AttributeError: Can't get attribute 'GenericResult[int]' on <module 'ccflow.result.generic' from '/Users/neej/dev/ccflow/ccflow/result/generic.py'>

=== cloudpickle protocol=5 ===
returncode: 1
stdout: has GenericResult[int] before unpickle: False
stderr: Traceback (most recent call last):
  File "<string>", line 11, in <module>
AttributeError: Can't get attribute 'GenericResult[int]' on <module 'ccflow.result.generic' from '/Users/neej/dev/ccflow/ccflow/result/generic.py'>
```

Ray worker output with `top_level_eval = True`:

```text
=== cloudpickle default protocol ===
worker has GenericResult[int] before unpickle: False
Traceback (most recent call last):
  File "/Users/neej/dev/ccflow/repro_generic_result_ray_cloudpickle.py", line 55, in load_payload
    obj = cloudpickle.loads(base64.b64decode(encoded_payload))
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
AttributeError: Can't get attribute 'GenericResult[int]' on <module 'ccflow.result.generic' from '/Users/neej/dev/ccflow/ccflow/result/generic.py'>

=== cloudpickle protocol=5 ===
worker has GenericResult[int] before unpickle: False
Traceback (most recent call last):
  File "/Users/neej/dev/ccflow/repro_generic_result_ray_cloudpickle.py", line 55, in load_payload
    obj = cloudpickle.loads(base64.b64decode(encoded_payload))
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
AttributeError: Can't get attribute 'GenericResult[int]' on <module 'ccflow.result.generic' from '/Users/neej/dev/ccflow/ccflow/result/generic.py'>
```

This reproduces with both cloudpickle's default protocol and `protocol=5`, so
the failure is not caused by protocol 5.

## Environment

- OS: macOS 26.2, build 25C56
- Python version: Python 3.12.12 (`uv run python --version`)
- Package version: ccflow 0.8.3
- cloudpickle: 3.1.2
- ray: 2.55.1
- pydantic: 2.13.4
- cclfow: main (commit: 17905fb8c27d8fb7656764c4db98aa20482a3e8a )

Package versions were collected with:

```bash
uv run python -c "import importlib.metadata as m; print('ccflow', m.version('ccflow')); print('cloudpickle', m.version('cloudpickle')); print('ray', m.version('ray')); print('pydantic', m.version('pydantic'))"
```

## Additional Context

Pydantic's generic specialization creation mutates module globals only in some
call contexts. This can be observed directly:

```python
import ccflow.result.generic as generic_module
from ccflow import GenericResult

assert not hasattr(generic_module, "GenericResult[int]")

cls = GenericResult[int]

assert hasattr(generic_module, "GenericResult[int]")
assert generic_module.__dict__["GenericResult[int]"] is cls
```

That side effect is process-local. A fresh Ray worker may import
`ccflow.result.generic` and `GenericResult` without ever evaluating
`GenericResult[int]`, so the module global expected by pickle does not exist.

The robust fix should avoid relying on Pydantic's process-local global
registration for ccflow generic result/context instances. Instead, generic
instances should be serialized by contract:

1. origin class, e.g. `GenericResult`;
2. generic args, e.g. `(int,)`;
3. Pydantic instance state.

Conceptually:

```python
def restore_generic_pydantic_instance(origin, args, state):
    cls = origin[args] if args else origin
    obj = cls.__new__(cls)
    obj.__setstate__(state)
    return obj
```

This should probably be fixed at the `ResultBase`/`ContextBase` serialization
layer.

One uv/Ray caveat from this repo: plain `uv run python
repro_generic_result_ray_cloudpickle.py` may cause Ray workers to bootstrap a
temporary uv environment that does not include `ray`, producing
`ModuleNotFoundError: No module named 'ray'` before the intended repro executes.
Using `uv run --active` from the repo venv reaches the actual cloudpickle
failure shown above.

NeejWeej:
# Pydantic Generic Pickle and Ray Notes

# TLDR

Concrete Pydantic generic `BaseModel` specializations can be fragile across
fresh-process pickle/cloudpickle boundaries. ccflow works around this for
concrete generic `ccflow.BaseModel` instances by pickling them as stable
`origin + args + state` data and recreating the specialized class during load.
Fixes #210.

## Summary

This issue is a mismatch between three things:

1. Python pickle prefers to reconstruct classes by importing a global module path.
2. Pydantic creates concrete generic model classes, such as `GenericResult[int]`, dynamically at runtime.
3. Ray workers are fresh processes and may unpickle an object before the same concrete generic class has been created in that worker.

The bug is not specific to `GenericResult`. Any concrete Pydantic generic `BaseModel` specialization can have the same problem if that specialized class object crosses a process boundary before the receiver has materialized it.

The fix is confusing because there are two separate objects involved:

- the model instance being pickled, such as `GenericResult[int](value=5)`
- the generated model class objects that may appear in the instance's class or
  generic type arguments, such as `GenericResult[int]`, `ListResult[int]`, or
  `CallableModelGenericType[NullContext, GenericResult[int]]`

Fixing only the top-level instance class is not enough if generated generic
classes are also embedded inside the type arguments that define that instance's
specialized class.

## What Pydantic Does

Pydantic v2 does not require `pydantic.generics.GenericModel`; normal `BaseModel` subclasses can be generic. When code evaluates:

```python
GenericResult[int]
```

Pydantic runs `BaseModel.__class_getitem__`. In the local environment, this is Pydantic `2.13.4`.
The relevant source is pinned to the `v2.13.4` tag here:
https://github.com/pydantic/pydantic/blob/v2.13.4/pydantic/main.py#L904-L969

The relevant flow is:

1. Check Pydantic's generic specialization cache.
2. Map type variables to concrete args.
3. Compute a parametrized model name, such as `GenericResult[int]`.
4. Call `_generics.create_generic_submodel(...)`.
5. Cache the generated class.
6. Return that generated class.

The generated class has metadata like:

```python
GenericResult[int].__pydantic_generic_metadata__
```

with shape:

```python
{
    "origin": GenericResult,
    "args": (int,),
    "parameters": (),
}
```

For module-level models like this repro, `origin` is stable and importable. In
general, the reducer still relies on the origin class itself being importable or
otherwise serializable by cloudpickle. The generated specialized class is
runtime-created.

In Pydantic's `_generics.create_generic_submodel`, the new subclass is created
with the origin model's `__module__` and generic metadata. The relevant source
is pinned here:
https://github.com/pydantic/pydantic/blob/v2.13.4/pydantic/_internal/_generics.py#L105-L149

```python
namespace = {"__module__": origin.__module__}
bases = (origin,)
...
created_model = meta(
    model_name,
    bases,
    namespace,
    __pydantic_generic_metadata__={
        "origin": origin,
        "args": args,
        "parameters": params,
    },
    ...
)
```

Then Pydantic conditionally registers the generated class in the origin module
when `_get_caller_frame_info(...)` decides the specialization was created from
a global context:
https://github.com/pydantic/pydantic/blob/v2.13.4/pydantic/_internal/_generics.py#L140-L147

```python
model_module, called_globally = _get_caller_frame_info(depth=3)
if called_globally:
    reference_module_globals = sys.modules[created_model.__module__].__dict__
    reference_module_globals.setdefault(reference_name, created_model)
```

That global registration is the key point. Sometimes a process has `ccflow.result.generic.GenericResult[int]` as a module attribute because that process already materialized it in a context Pydantic considers global. A fresh process may not.

## What Pickle and Cloudpickle Do

Pickle generally reconstructs class objects by global reference:

```text
module name + class name
```

For an ordinary class, this is fine:

```python
ccflow.result.generic.GenericResult
```

can be imported in any process.

For a generated specialization, pickle/cloudpickle may see:

```python
__module__ = "ccflow.result.generic"
__name__ = "GenericResult[int]"
```

and serialize it by reference as if this were importable:

```python
ccflow.result.generic.GenericResult[int]
```

That works in the process that created and registered the class. It can fail in a fresh process:

```text
AttributeError: Can't get attribute 'GenericResult[int]' on module 'ccflow.result.generic'
```

Ray makes this easy to hit because Ray workers are separate Python processes. Importing `GenericResult` in the worker does not necessarily create `GenericResult[int]`. If unpickling happens first, the generated class name is missing.

## Pydantic's Instance Pickle Behavior

Pydantic already has instance pickle machinery.

`BaseModel.__getstate__()` returns a dict containing Pydantic's internal model
state, and `BaseModel.__setstate__()` restores that state directly. The source
is pinned here:
https://github.com/pydantic/pydantic/blob/v2.13.4/pydantic/main.py#L1145-L1160

```python
{
    "__dict__": self.__dict__,
    "__pydantic_extra__": self.__pydantic_extra__,
    "__pydantic_fields_set__": self.__pydantic_fields_set__,
    "__pydantic_private__": private,
}
```

That is important because pickle should preserve an already-validated object. It should not rerun normal validation, coerce values again, drop private attrs, or rebuild the object through `model_validate`.

ccflow already overrides `__getstate__` / `__setstate__` slightly to make `__pydantic_fields_set__` deterministic in pickle output.

The new fix keeps this Pydantic state-based behavior. For concrete generic
specializations of `ccflow.BaseModel`, it changes only the reduce recipe.

## The Exact Failure

A minimal failing shape is:

```python
payload = cloudpickle.dumps(GenericResult[int](value=5))
```

Then in a fresh process that has imported `GenericResult` but has not evaluated `GenericResult[int]`:

```python
cloudpickle.loads(payload)
```

can fail because the receiver has the origin class:

```python
ccflow.result.generic.GenericResult
```

but not the generated specialization:

```python
ccflow.result.generic.GenericResult[int]
```

The problem is broader than the top-level class:

```python
GenericResult[ListResult[int]](...)
```

Here the top-level class is `GenericResult[ListResult[int]]`, and the generic arg contains another generated class, `ListResult[int]`.

This also appears inside typing aliases:

```python
GenericResult[list[ListResult[int]]](...)
GenericResult[typing.List[ListResult[int]]](...)
GenericResult[typing.Callable[[ListResult[int]], int]](...)
GenericResult[GenericContext[int] | None](...)
```

`typing.Callable` is especially annoying because its parameter types can appear as a plain Python list inside `typing.get_args()`:

```python
typing.get_args(Callable[[ListResult[int]], int])
# ([ListResult[int]], int)
```

So a helper that only walks `typing.get_args()` recursively can still miss generated classes inside that list.

Generated classes can also appear as field values:

```python
GenericResult[type](value=ListResult[int])
```

Even if the instance class is reconstructed correctly, the field value
`ListResult[int]` would otherwise be pickled by its fragile generated class
name. That field-value shape is not fixed by the current change. The current
fix deliberately covers generated classes used as the instance class and inside
generic type arguments, while leaving arbitrary class objects stored in model
state to pickle/cloudpickle's normal behavior.

## Pydantic-Only Repro

The smallest useful repro does not need ccflow or Ray. It only needs a plain
Pydantic generic model, two Python processes, and a cold receiver that imports
the generic origin class without first materializing the concrete specialization.

Consider the following repro:

```python
# /// script
# dependencies = [
#   "cloudpickle",
#   "pydantic>=2.6,<3",
# ]
# ///
"""Minimal Pydantic-only repro for generic BaseModel pickle lookup failures.

This intentionally does not import ccflow. It creates a normal importable module
containing a plain Pydantic generic model, pickles ``Box[int](...)`` in one
process, then tries to unpickle it in a second fresh process that imported
``Box`` but did not materialize ``Box[int]``.
"""

from __future__ import annotations

import base64
import subprocess
import sys
import tempfile
import textwrap
from pathlib import Path


def run_python(script: str) -> subprocess.CompletedProcess[str]:
    return subprocess.run(
        [sys.executable, "-c", textwrap.dedent(script)],
        capture_output=True,
        text=True,
        timeout=30,
    )


with tempfile.TemporaryDirectory() as temp_dir:
    root = Path(temp_dir)
    module_path = root / "pydantic_generic_model.py"
    module_path.write_text(
        textwrap.dedent(
            """
            from typing import Generic, TypeVar

            from pydantic import BaseModel

            T = TypeVar("T")

            class Box(BaseModel, Generic[T]):
                value: T
            """
        )
    )

    creator = run_python(
        f"""
        import base64
        import cloudpickle
        import pydantic
        import sys

        sys.path.insert(0, {str(root)!r})
        import pydantic_generic_model
        from pydantic_generic_model import Box

        print("pydantic_version=", pydantic.__version__)
        value = Box[int](value=5)
        print("creator_has_Box_int=", hasattr(pydantic_generic_model, "Box[int]"))
        print(base64.b64encode(cloudpickle.dumps(value, protocol=5)).decode())
        """
    )
    if creator.returncode != 0:
        raise SystemExit(creator.stderr)

    payload = creator.stdout.splitlines()[-1]
    print("=== creator ===")
    print("\n".join(creator.stdout.splitlines()[:-1]))

    cold_loader = run_python(
        f"""
        import base64
        import cloudpickle
        import sys

        sys.path.insert(0, {str(root)!r})
        import pydantic_generic_model
        from pydantic_generic_model import Box

        print("loader_has_Box_int_before=", hasattr(pydantic_generic_model, "Box[int]"))
        value = cloudpickle.loads(base64.b64decode({payload!r}))
        print(value)
        """
    )
    print("\n=== cold receiver: imports Box but not Box[int] ===")
    print("returncode:", cold_loader.returncode)
    print("stdout:")
    print(cold_loader.stdout.rstrip() or "<empty>")
    print("stderr:")
    print(cold_loader.stderr.rstrip() or "<empty>")

    warm_loader = run_python(
        f"""
        import base64
        import cloudpickle
        import sys

        sys.path.insert(0, {str(root)!r})
        import pydantic_generic_model
        from pydantic_generic_model import Box

        _ = Box[int]
        print("loader_has_Box_int_before=", hasattr(pydantic_generic_model, "Box[int]"))
        value = cloudpickle.loads(base64.b64decode({payload!r}))
        print(value)
        """
    )
    print("\n=== warm receiver: materializes Box[int] before load ===")
    print("returncode:", warm_loader.returncode)
    print("stdout:")
    print(warm_loader.stdout.rstrip() or "<empty>")
    print("stderr:")
    print(warm_loader.stderr.rstrip() or "<empty>")

```

Then it runs three subprocess steps:

1. A creator process imports `Box`, evaluates `Box[int]`, constructs
   `Box[int](value=5)`, and serializes it with `cloudpickle`.
2. A cold receiver process imports `Box` but does not evaluate `Box[int]`
   before calling `cloudpickle.loads(...)`.
3. A warm receiver process imports `Box`, evaluates `Box[int]`, then calls
   `cloudpickle.loads(...)`.

The observed output is:

```text
=== creator ===
pydantic_version= 2.13.4
creator_has_Box_int= True

=== cold receiver: imports Box but not Box[int] ===
returncode: 1
stdout:
loader_has_Box_int_before= False
stderr:
Traceback (most recent call last):
  File "<string>", line 11, in <module>
AttributeError: Can't get attribute 'Box[int]' on <module 'pydantic_generic_model' ...>

=== warm receiver: materializes Box[int] before load ===
returncode: 0
stdout:
loader_has_Box_int_before= True
value=5
stderr:
<empty>
```

That is the core bug in isolation. The creating process has a generated
`Box[int]` class registered on the module. The cold receiving process has only
the importable generic origin class `Box`, so pickle's global lookup for
`Box[int]` fails. The warm receiver evaluates `Box[int]` at module/global scope
before unpickling, which causes Pydantic to install the same generated class as
a module attribute before pickle tries to resolve it. That proves this is an
import/materialization-ordering problem rather than a ccflow model-definition
problem.

The same repro was checked with several Pydantic 2.x releases using:

```bash
uv run --with pydantic==<version> <pydantic-only-repro-script>
```

| Pydantic version | Cold receiver result | Warm receiver result |
| --- | --- | --- |
| 2.6.4 | fails with missing `Box[int]` module attribute | succeeds |
| 2.7.4 | fails with missing `Box[int]` module attribute | succeeds |
| 2.8.2 | fails with missing `Box[int]` module attribute | succeeds |
| 2.9.2 | fails with missing `Box[int]` module attribute | succeeds |
| 2.10.6 | fails with missing `Box[int]` module attribute | succeeds |
| 2.11.10 | fails with missing `Box[int]` module attribute | succeeds |
| 2.12.5 | fails with missing `Box[int]` module attribute | succeeds |
| 2.13.4 | fails with missing `Box[int]` module attribute | succeeds |

This is not a regression in a recent Pydantic minor release. The behavior is
stable across the tested 2.x line.

## Why the Fix Uses `__reduce_ex__`

`__reduce_ex__` is the pickle hook that returns a reconstruction recipe.

For concrete generic specializations of `ccflow.BaseModel`, ccflow now returns
a recipe like:

```python
(
    _new_ccflow_generic_model,
    (origin, portable_args),
    pydantic_state,
)
```

For:

```python
GenericResult[int](value=5)
```

the recipe is conceptually:

```python
origin = GenericResult
args = (int,)
state = self.__getstate__()
```

On load, pickle first calls the reducer function:

```python
cls = GenericResult[int]
obj = cls.__new__(cls)
```

Then, because the reducer returned `pydantic_state` as the third tuple element,
pickle applies that state to the object:

```python
obj.__setstate__(pydantic_state)
```

This uses Pydantic's own generic construction path in the receiving process,
while still letting pickle apply Pydantic's normal state protocol. The receiver
does not need to already have a global `GenericResult[int]` module attribute.

## Why Type Arguments Need Special Handling

The generic argument portability layer is the ugly part, but it is solving a
real second-order problem.

If we only serialize the top-level class as:

```python
origin + args
```

then this works:

```python
GenericResult[int](value=5)
```

but these can still fail:

```python
GenericResult[ListResult[int]](...)
GenericResult[list[ListResult[int]]](...)
GenericResult[Callable[[ListResult[int]], int]](...)
```

because `ListResult[int]` is itself a generated Pydantic generic class.

The helper therefore handles:

- generic type arguments
- typing aliases that contain generic type arguments
- `list` and `tuple` containers that can appear inside type expressions, such
  as the callable parameter list

The raw Pydantic state remains in the outer pickle stream. That is intentional:
pickle keeps its own memo table there, which is required to preserve shared
references, cycles, and protocol-5 buffers. The reducer only changes how the
generic class is recreated; it does not recursively rewrite arbitrary model
field/private state.

It intentionally does **not** treat model instances as type specs. A value like:

```python
ListResult[int](value=[1])
```

is still a model instance and should be pickled as an instance. Its own `__reduce_ex__` will handle its generated class. Accidentally converting instances into class specs would corrupt data.

## Blast Radius

There are two related guards:

```python
isinstance(value, type)
and value.__pydantic_generic_metadata__["origin"] is not None
```

That predicate identifies concrete generated Pydantic generic classes such as:

```python
GenericResult[int]
ListResult[str]
CallableModelGenericType[NullContext, GenericResult[int]]
```

The `BaseModel.__reduce_ex__` override runs when pickling a ccflow model
instance whose `type(self)` satisfies that predicate. So it **does** apply to:

```python
GenericResult[int](value=5)
```

It does **not** apply to:

```python
GenericResult
ParentModel
ParentModel(...)
```

Normal non-generic `BaseModel` instances continue using the default reducer
path, plus ccflow's existing deterministic `__getstate__` / `__setstate__`
hooks.

The custom reduce recipe is created only during pickling of concrete generic
ccflow `BaseModel` instances. It does not run during:

- normal construction
- validation
- `model_dump`
- equality
- compute/call paths
- registry lookup

The performance cost is therefore limited to pickling generic model instances.
The extra work is walking the generic type arguments for the model class. The
actual Pydantic instance state is still handled by the surrounding pickle
operation.

### Why not call `model_validate` on restore?

Because pickle should restore object state, not validate new input.

Revalidation can:

- coerce values differently
- drop private attrs
- rerun validators with side effects
- fail if the current schema changed
- fail if the object was valid when created but no longer validates under newer code

Pydantic's own pickle support uses `__getstate__` / `__setstate__`, so the ccflow fix follows that model.

### Why not rely on cloudpickle to serialize the generated class by value?

Sometimes cloudpickle can serialize dynamic classes by value. But Pydantic specializations are not just ordinary dynamic classes. They carry generated schemas, validators, serializers, generic metadata, and cache behavior.

Also, if the class appears to be importable by module/name in the creating process, cloudpickle can choose a global-reference path. That is exactly the fragile path that fails in a fresh receiver.

The stable representation for a Pydantic generic specialization is not the generated class object. It is:

```python
origin class + generic args
```

### Why not globally register every generated specialization?

Pydantic already conditionally registers generated specializations when it thinks they were created globally. But a fresh Ray worker has not necessarily executed the same specialization expression yet.

Trying to eagerly register all possible specializations is impossible. Registering during serialization still would not help the receiver unless the receiver imports side effects in the same order.

### Why not monkeypatch pickle/cloudpickle for all classes?

Generated Pydantic specializations are class objects, so a global reducer would mean changing behavior for `type` or for broad classes of model classes. That is much wider than this bug.

The ccflow fix keeps the custom behavior inside ccflow `BaseModel` instance pickling.

### Why not support fields containing generated classes too?

That is a real broader issue, but fixing it at the state-value level is a much
bigger change. It requires intercepting arbitrary class objects inside the
pickle stream or walking Pydantic state manually, both of which can disturb
pickle's normal identity/cycle/buffer semantics if done carelessly.

The current fix chooses the smaller and safer boundary: generated classes that
define the model instance's own type, plus generated classes inside that type's
generic arguments. A field value like `GenericResult[type](value=ListResult[int])`
can still fail in a cold receiver and remains out of scope.

## How Broad Is This Problem?

It affects concrete Pydantic generic specializations crossing a process
boundary by pickle/cloudpickle when the receiver has not already materialized
the same specialization.

Shapes this PR fixes:

- instance class: `GenericResult[int](...)`
- nested generic arg: `GenericResult[ListResult[int]](...)`
- builtin alias arg: `GenericResult[list[ListResult[int]]](...)`
- dict alias arg: `GenericResult[dict[str, GenericContext[int]]](...)`
- `typing` alias arg: `GenericResult[typing.List[ListResult[int]]](...)`
- callable alias arg: `GenericResult[typing.Callable[[ListResult[int]], int]](...)`
- single-argument special-form args accepted by Pydantic in this context, such
  as `typing.ClassVar[ListResult[int]]` and `typing.Final[ListResult[int]]`
- union arg: `GenericResult[GenericContext[int] | None](...)`
- optional alias arg: `GenericResult[typing.Optional[ListResult[int]]](...)`

Not affected:

- non-generic `BaseModel` classes
- generic origin classes before parametrization, such as `GenericResult`
- normal runtime use that does not pickle

Handled by normal nested pickling, not by rewriting field/private state:

- generic ccflow model instances stored as field values; pickle visits those
  instances normally, and each instance's own reducer handles its generated
  class

Helper-level coverage only:

- `CallableModelGenericType[NullContext, GenericResult[int]]` as a type
  argument
- `typing.Required[ListResult[int]]` and
  `typing.NotRequired[ListResult[int]]`; Pydantic rejects these as
  `GenericResult[...]` arguments in this context, but the restore helper
  handles the one-argument special-form shape

Known not fixed:

- class-valued state such as `GenericResult[type](value=ListResult[int])`
- arbitrary metadata containers inside type expressions, such as
  `Annotated[int, frozenset([ListResult[int]])]`; the helper walks normal
  typing args plus explicit `list`/`tuple` containers, not every object that
  can be embedded in metadata

## Why This Is Hard To Read

The code is confusing because it has to preserve three different pieces of
pickle/type state:

1. Pydantic model instance state: restored through `__getstate__` / `__setstate__`
2. Pydantic generic class identity: restored through `origin[args]`
3. Python typing alias form: rebuilt as the same kind of type expression, such
   as builtin `list[...]`, `typing.List[...]`, `typing.Optional[...]`,
   `typing.Callable[...]`, or a PEP 604 `A | B` union

It also has to avoid a dangerous false positive:

```python
ListResult[int]          # generated class: make portable
ListResult[int](...)     # model instance: do not convert to a class spec
```

That is why there are separate helpers for:

- detecting concrete Pydantic generic specialization classes
- making generic type arguments portable
- restoring generic type arguments
- allocating the rebuilt generic model instance before pickle applies
  Pydantic's normal `__setstate__`

The resulting implementation is not aesthetically simple, but each piece exists to handle a real pickle/Ray failure mode.

## References

- Pydantic dynamic model docs. Pydantic explicitly notes that dynamically created models must be globally defined and have `__module__` provided to be pickleable:
  https://docs.pydantic.dev/latest/concepts/models/#dynamic-model-creation

- Pydantic `BaseModel.__class_getitem__` source for generic specialization:
  https://github.com/pydantic/pydantic/blob/v2.13.4/pydantic/main.py#L904-L969

- Pydantic `_generics.create_generic_submodel` source for dynamic generated
  generic subclasses and conditional module registration:
  https://github.com/pydantic/pydantic/blob/v2.13.4/pydantic/_internal/_generics.py#L105-L149

- Pydantic `BaseModel.__getstate__` / `BaseModel.__setstate__` source showing
  state-based pickle behavior:
  https://github.com/pydantic/pydantic/blob/v2.13.4/pydantic/main.py#L1145-L1160

- Pydantic issue #9668, broad background on Python compatibility work that mentions generic models and pickling among affected areas. It is not this exact bug:
  https://github.com/pydantic/pydantic/issues/9668

- Prefect's `add_cloudpickle_reduction`, an analogous downstream precedent for adding reducer logic around Pydantic model classes in workflow/distributed execution contexts:
  https://reference.prefect.io/prefect/utilities/pydantic/#add_cloudpickle_reduction

### DIFF
--- a/ccflow/base.py
+++ b/ccflow/base.py
@@ -7,28 +7,9 @@ import logging
 import pathlib
 import platform
 import sys
-import types
-import typing
 import warnings
-from functools import singledispatch
 from types import GenericAlias, MappingProxyType
-from typing import (
-    TYPE_CHECKING,
-    Any,
-    Callable,
-    ClassVar,
-    Dict,
-    Generic,
-    List,
-    NamedTuple,
-    Optional,
-    Tuple,
-    Type,
-    TypeVar,
-    Union,
-    get_args,
-    get_origin,
-)
+from typing import TYPE_CHECKING, Any, Callable, ClassVar, Dict, Generic, List, Optional, Tuple, Type, TypeVar, Union, get_args, get_origin
 
 import pydantic
 from packaging import version
@@ -52,6 +33,7 @@ from typing_extensions import Self
 
 from .exttypes.pyobjectpath import PyObjectPath
 from .local_persistence import register_ccflow_import_path, sync_to_module
+from .pickling import reduce_generic_model_instance
 
 log = logging.getLogger(__name__)
 
@@ -197,155 +179,6 @@ class _SerializeAsAnyMeta(ModelMetaclass):
         namespaces["__annotations__"] = annotations
 
         return super().__new__(self, name, bases, namespaces, **kwargs)
-
-
-class _PydanticGenericTypeSpec(NamedTuple):
-    # Pickle must not rely on process-local names like ``GenericResult[int]``.
-    # Store generic args as data so the receiver can rebuild ``origin[args]``.
-    origin: Type[PydanticBaseModel]
-    args: Tuple[Any, ...]
-
-
-class _GenericAliasTypeSpec(NamedTuple):
-    # Builtin/typing aliases such as ``list[GenericResult[int]]`` are normally
-    # pickleable, but their nested Pydantic generic args may not be. Only wrap
-    # aliases when at least one nested arg had to become portable data.
-    origin: Any
-    args: Tuple[Any, ...]
-    typing_name: Optional[str]
-
-
-def _is_pydantic_generic_specialization(value: Any) -> bool:
-    if not isinstance(value, type):
-        return False
-    metadata = getattr(value, "__pydantic_generic_metadata__", None)
-    return bool(metadata and metadata.get("origin") is not None)
-
-
-@singledispatch
-def _portable_generic_type_arg(value: Any) -> Any:
-    """Convert fragile generic type arguments into pickleable rebuild specs.
-
-    The top-level reducer stores ``origin`` and ``args`` for a model instance,
-    for example ``GenericResult`` and ``(int,)`` for ``GenericResult[int]``.
-    Some args are themselves runtime Pydantic generic classes, such as the
-    ``ListResult[int]`` in ``GenericResult[ListResult[int]]``. Those nested
-    classes have the same cross-process problem as the top-level class: pickle
-    may look for a process-local global named ``ListResult[int]``.
-
-    This helper walks only type arguments, not model field values. Plain
-    importable args like ``int`` are left alone. Pydantic generic
-    specializations become explicit ``(origin, args)`` specs. Generic aliases
-    like ``list[ListResult[int]]`` are wrapped only if one of their nested args
-    needed that conversion. Some type args contain list/tuple containers, such
-    as the argument list in ``typing.Callable[[ListResult[int]], int]``; those
-    containers are also walked because they are part of the type expression.
-    """
-    if _is_pydantic_generic_specialization(value):
-        # Example: ``ListResult[int]`` becomes
-        # ``_PydanticGenericTypeSpec(ListResult, (int,))``. This avoids storing
-        # the generated class object that may not be globally registered in the
-        # receiving process.
-        metadata = value.__pydantic_generic_metadata__
-        return _PydanticGenericTypeSpec(
-            metadata["origin"],
-            tuple(_portable_generic_type_arg(arg) for arg in metadata.get("args", ())),
-        )
-
-    origin = get_origin(value)
-    args = get_args(value)
-    if origin is not None and args:
-        # Example: ``list[ListResult[int]]`` is a normal generic alias, but it
-        # contains a fragile Pydantic specialization. Rebuild the alias from a
-        # portable version of its args only when recursion changed something.
-        portable_args = tuple(_portable_generic_type_arg(arg) for arg in args)
-        if portable_args != args:
-            typing_name = getattr(value, "_name", None) if getattr(value, "__module__", None) == "typing" else None
-            return _GenericAliasTypeSpec(origin, portable_args, typing_name)
-    return value
-
-
-@_portable_generic_type_arg.register(list)
-def _(value: list) -> Any:
-    portable_items = [_portable_generic_type_arg(item) for item in value]
-    return portable_items if portable_items != value else value
-
-
-@_portable_generic_type_arg.register(tuple)
-def _(value: tuple) -> Any:
-    portable_items = tuple(_portable_generic_type_arg(item) for item in value)
-    return portable_items if portable_items != value else value
-
-
-@singledispatch
-def _restore_generic_type_arg(value: Any) -> Any:
-    return value
-
-
-@_restore_generic_type_arg.register(_PydanticGenericTypeSpec)
-def _(value: _PydanticGenericTypeSpec) -> Any:
-    origin, args = value
-    return origin[tuple(_restore_generic_type_arg(arg) for arg in args)]
-
-
-@_restore_generic_type_arg.register(_GenericAliasTypeSpec)
-def _(value: _GenericAliasTypeSpec) -> Any:
-    origin, args, typing_name = value
-    restored_args = tuple(_restore_generic_type_arg(arg) for arg in args)
-    if typing_name is not None:
-        alias = getattr(typing, typing_name)
-        if typing_name == "Optional":
-            non_none_args = tuple(arg for arg in restored_args if arg is not type(None))
-            if len(non_none_args) == 1:
-                return alias[non_none_args[0]]
-        return alias[restored_args]
-    try:
-        return origin[restored_args]
-    except TypeError as exc:
-        if len(restored_args) == 1:
-            try:
-                return origin[restored_args[0]]
-            except TypeError:
-                pass
-        # ``types.UnionType`` is not itself subscriptable; rebuild PEP 604
-        # unions from their members if one of those members needed
-        # portable restoration.
-        union_type = getattr(types, "UnionType", None)
-        if union_type is not None and origin is union_type:
-            result = restored_args[0]
-            for arg in restored_args[1:]:
-                result = result | arg
-            return result
-        raise exc
-
-
-@_restore_generic_type_arg.register(list)
-def _(value: list) -> list:
-    return [_restore_generic_type_arg(item) for item in value]
-
-
-@_restore_generic_type_arg.register(tuple)
-def _(value: tuple) -> tuple:
-    return tuple(_restore_generic_type_arg(item) for item in value)
-
-
-def _new_ccflow_generic_model(origin: Type[PydanticBaseModel], args: Tuple[Any, ...]) -> PydanticBaseModel:
-    """Restore a Pydantic generic specialization without a process-local global.
-
-    Pydantic registers runtime generic specializations like ``GenericResult[int]``
-    in the origin module only in the process that first materializes them. Pickle
-    can then serialize instances by that global name, but fresh workers may not
-    have created the same specialization yet. Restore from the stable origin
-    class plus generic args instead of relying on that process-local module
-    mutation.
-    """
-
-    # Materialize the specialized class in this process. Pickle will restore the
-    # raw Pydantic state afterwards through BaseModel.__setstate__; keeping that
-    # state in the outer pickle stream preserves memo semantics for shared
-    # references, cycles, and protocol-5 buffers.
-    cls = origin[tuple(_restore_generic_type_arg(arg) for arg in args)]
-    return cls.__new__(cls)
 
 
 _BASE_MODEL_METACLASS = ModelMetaclass if _USE_RUNTIME_POLYMORPHIC_SERIALIZATION else _SerializeAsAnyMeta
@@ -522,18 +355,9 @@ class BaseModel(PydanticBaseModel, _RegistryMixin, metaclass=_BASE_MODEL_METACLA
         super().__setstate__(state)
 
     def __reduce_ex__(self, protocol):
-        if _is_pydantic_generic_specialization(type(self)):
-            # Pydantic's default reducer may serialize runtime generic classes
-            # by names like ``GenericResult[int]``. Those names exist only
-            # after a process has materialized that exact specialization, so
-            # Ray/fresh workers can fail to import them while unpickling.
-            # Serialize a stable constructor instead: generic origin plus
-            # portable generic args. Leave Pydantic instance state to the outer
-            # pickle stream so shared references, cycles, and protocol-5 buffers
-            # keep normal pickle semantics.
-            metadata = type(self).__pydantic_generic_metadata__
-            args = tuple(_portable_generic_type_arg(arg) for arg in metadata.get("args", ()))
-            return (_new_ccflow_generic_model, (metadata["origin"], args), self.__getstate__())
+        reducer = reduce_generic_model_instance(self)
+        if reducer is not None:
+            return reducer
         return super().__reduce_ex__(protocol)
 
 
--- a/ccflow/base.py
+++ b/ccflow/base.py
@@ -7,9 +7,28 @@ import logging
 import pathlib
 import platform
 import sys
+import types
+import typing
 import warnings
+from functools import singledispatch
 from types import GenericAlias, MappingProxyType
-from typing import TYPE_CHECKING, Any, Callable, ClassVar, Dict, Generic, List, Optional, Tuple, Type, TypeVar, Union, get_args, get_origin
+from typing import (
+    TYPE_CHECKING,
+    Any,
+    Callable,
+    ClassVar,
+    Dict,
+    Generic,
+    List,
+    NamedTuple,
+    Optional,
+    Tuple,
+    Type,
+    TypeVar,
+    Union,
+    get_args,
+    get_origin,
+)
 
 from packaging import version
 
@@ -163,6 +182,155 @@ class _SerializeAsAnyMeta(ModelMetaclass):
         namespaces["__annotations__"] = annotations
 
         return super().__new__(self, name, bases, namespaces, **kwargs)
+
+
+class _PydanticGenericTypeSpec(NamedTuple):
+    # Pickle must not rely on process-local names like ``GenericResult[int]``.
+    # Store generic args as data so the receiver can rebuild ``origin[args]``.
+    origin: Type[PydanticBaseModel]
+    args: Tuple[Any, ...]
+
+
+class _GenericAliasTypeSpec(NamedTuple):
+    # Builtin/typing aliases such as ``list[GenericResult[int]]`` are normally
+    # pickleable, but their nested Pydantic generic args may not be. Only wrap
+    # aliases when at least one nested arg had to become portable data.
+    origin: Any
+    args: Tuple[Any, ...]
+    typing_name: Optional[str]
+
+
+def _is_pydantic_generic_specialization(value: Any) -> bool:
+    if not isinstance(value, type):
+        return False
+    metadata = getattr(value, "__pydantic_generic_metadata__", None)
+    return bool(metadata and metadata.get("origin") is not None)
+
+
+@singledispatch
+def _portable_generic_type_arg(value: Any) -> Any:
+    """Convert fragile generic type arguments into pickleable rebuild specs.
+
+    The top-level reducer stores ``origin`` and ``args`` for a model instance,
+    for example ``GenericResult`` and ``(int,)`` for ``GenericResult[int]``.
+    Some args are themselves runtime Pydantic generic classes, such as the
+    ``ListResult[int]`` in ``GenericResult[ListResult[int]]``. Those nested
+    classes have the same cross-process problem as the top-level class: pickle
+    may look for a process-local global named ``ListResult[int]``.
+
+    This helper walks only type arguments, not model field values. Plain
+    importable args like ``int`` are left alone. Pydantic generic
+    specializations become explicit ``(origin, args)`` specs. Generic aliases
+    like ``list[ListResult[int]]`` are wrapped only if one of their nested args
+    needed that conversion. Some type args contain list/tuple containers, such
+    as the argument list in ``typing.Callable[[ListResult[int]], int]``; those
+    containers are also walked because they are part of the type expression.
+    """
+    if _is_pydantic_generic_specialization(value):
+        # Example: ``ListResult[int]`` becomes
+        # ``_PydanticGenericTypeSpec(ListResult, (int,))``. This avoids storing
+        # the generated class object that may not be globally registered in the
+        # receiving process.
+        metadata = value.__pydantic_generic_metadata__
+        return _PydanticGenericTypeSpec(
+            metadata["origin"],
+            tuple(_portable_generic_type_arg(arg) for arg in metadata.get("args", ())),
+        )
+
+    origin = get_origin(value)
+    args = get_args(value)
+    if origin is not None and args:
+        # Example: ``list[ListResult[int]]`` is a normal generic alias, but it
+        # contains a fragile Pydantic specialization. Rebuild the alias from a
+        # portable version of its args only when recursion changed something.
+        portable_args = tuple(_portable_generic_type_arg(arg) for arg in args)
+        if portable_args != args:
+            typing_name = getattr(value, "_name", None) if getattr(value, "__module__", None) == "typing" else None
+            return _GenericAliasTypeSpec(origin, portable_args, typing_name)
+    return value
+
+
+@_portable_generic_type_arg.register(list)
+def _(value: list) -> Any:
+    portable_items = [_portable_generic_type_arg(item) for item in value]
+    return portable_items if portable_items != value else value
+
+
+@_portable_generic_type_arg.register(tuple)
+def _(value: tuple) -> Any:
+    portable_items = tuple(_portable_generic_type_arg(item) for item in value)
+    return portable_items if portable_items != value else value
+
+
+@singledispatch
+def _restore_generic_type_arg(value: Any) -> Any:
+    return value
+
+
+@_restore_generic_type_arg.register(_PydanticGenericTypeSpec)
+def _(value: _PydanticGenericTypeSpec) -> Any:
+    origin, args = value
+    return origin[tuple(_restore_generic_type_arg(arg) for arg in args)]
+
+
+@_restore_generic_type_arg.register(_GenericAliasTypeSpec)
+def _(value: _GenericAliasTypeSpec) -> Any:
+    origin, args, typing_name = value
+    restored_args = tuple(_restore_generic_type_arg(arg) for arg in args)
+    if typing_name is not None:
+        alias = getattr(typing, typing_name)
+        if typing_name == "Optional":
+            non_none_args = tuple(arg for arg in restored_args if arg is not type(None))
+            if len(non_none_args) == 1:
+                return alias[non_none_args[0]]
+        return alias[restored_args]
+    try:
+        return origin[restored_args]
+    except TypeError as exc:
+        if len(restored_args) == 1:
+            try:
+                return origin[restored_args[0]]
+            except TypeError:
+                pass
+        # ``types.UnionType`` is not itself subscriptable; rebuild PEP 604
+        # unions from their members if one of those members needed
+        # portable restoration.
+        union_type = getattr(types, "UnionType", None)
+        if union_type is not None and origin is union_type:
+            result = restored_args[0]
+            for arg in restored_args[1:]:
+                result = result | arg
+            return result
+        raise exc
+
+
+@_restore_generic_type_arg.register(list)
+def _(value: list) -> list:
+    return [_restore_generic_type_arg(item) for item in value]
+
+
+@_restore_generic_type_arg.register(tuple)
+def _(value: tuple) -> tuple:
+    return tuple(_restore_generic_type_arg(item) for item in value)
+
+
+def _new_ccflow_generic_model(origin: Type[PydanticBaseModel], args: Tuple[Any, ...]) -> PydanticBaseModel:
+    """Restore a Pydantic generic specialization without a process-local global.
+
+    Pydantic registers runtime generic specializations like ``GenericResult[int]``
+    in the origin module only in the process that first materializes them. Pickle
+    can then serialize instances by that global name, but fresh workers may not
+    have created the same specialization yet. Restore from the stable origin
+    class plus generic args instead of relying on that process-local module
+    mutation.
+    """
+
+    # Materialize the specialized class in this process. Pickle will restore the
+    # raw Pydantic state afterwards through BaseModel.__setstate__; keeping that
+    # state in the outer pickle stream preserves memo semantics for shared
+    # references, cycles, and protocol-5 buffers.
+    cls = origin[tuple(_restore_generic_type_arg(arg) for arg in args)]
+    return cls.__new__(cls)
 
 
 class BaseModel(PydanticBaseModel, _RegistryMixin, metaclass=_SerializeAsAnyMeta):
@@ -333,6 +501,21 @@ class BaseModel(PydanticBaseModel, _RegistryMixin, metaclass=_SerializeAsAnyMeta
     def __setstate__(self, state):
         state["__pydantic_fields_set__"] = set(state["__pydantic_fields_set__"])
         super().__setstate__(state)
+
+    def __reduce_ex__(self, protocol):
+        if _is_pydantic_generic_specialization(type(self)):
+            # Pydantic's default reducer may serialize runtime generic classes
+            # by names like ``GenericResult[int]``. Those names exist only
+            # after a process has materialized that exact specialization, so
+            # Ray/fresh workers can fail to import them while unpickling.
+            # Serialize a stable constructor instead: generic origin plus
+            # portable generic args. Leave Pydantic instance state to the outer
+            # pickle stream so shared references, cycles, and protocol-5 buffers
+            # keep normal pickle semantics.
+            metadata = type(self).__pydantic_generic_metadata__
+            args = tuple(_portable_generic_type_arg(arg) for arg in metadata.get("args", ()))
+            return (_new_ccflow_generic_model, (metadata["origin"], args), self.__getstate__())
+        return super().__reduce_ex__(protocol)
 
 
 class _ModelRegistryData(PydanticBaseModel):
--- a/ccflow/pickling.py
+++ b/ccflow/pickling.py
@@ -1,0 +1,140 @@
+"""Pickle helpers for ccflow model classes.
+
+This module works around Pydantic generic specializations sometimes pickling by
+process-local generated class names, such as ``GenericResult[int]``. Fresh
+processes may not have materialized those names yet, so unpickling can fail
+before ccflow code gets control. See the upstream Pydantic issue for the
+underlying behavior:
+
+https://github.com/pydantic/pydantic/issues/9390
+
+If Pydantic or cloudpickle eventually make generated generic model
+specializations portable across fresh processes without relying on module-global
+side effects, this module can likely be deleted and ``BaseModel.__reduce_ex__``
+can fall back to Pydantic's default reducer.
+"""
+
+import types
+from functools import singledispatch
+from typing import Any, NamedTuple, Optional, Tuple, Type, get_args, get_origin
+
+from pydantic import BaseModel as PydanticBaseModel
+
+__all__ = ("reduce_generic_model_instance",)
+
+
+class _GenericTypeSpec(NamedTuple):
+    # Pickle must not rely on process-local names like ``GenericResult[int]``.
+    # Store generic args as data so the receiver can rebuild ``origin[args]``.
+    origin: Any
+    args: Tuple[Any, ...]
+
+
+def _is_pydantic_generic_specialization(value: Any) -> bool:
+    if not isinstance(value, type):
+        return False
+    metadata = getattr(value, "__pydantic_generic_metadata__", None)
+    return bool(metadata and metadata.get("origin") is not None)
+
+
+@singledispatch
+def _portable_generic_type_arg_with_changed(value: Any) -> tuple[Any, bool]:
+    if _is_pydantic_generic_specialization(value):
+        metadata = value.__pydantic_generic_metadata__
+        portable_args, _ = _portable_generic_type_args(metadata["args"])
+        return _GenericTypeSpec(metadata["origin"], portable_args), True
+
+    origin = get_origin(value)
+    args = get_args(value)
+    if origin is not None and args:
+        portable_args, changed = _portable_generic_type_args(args)
+        if changed:
+            return _GenericTypeSpec(origin, portable_args), True
+    return value, False
+
+
+@_portable_generic_type_arg_with_changed.register(list)
+def _(value: list) -> tuple[Any, bool]:
+    portable_items = []
+    changed = False
+    for item in value:
+        portable_item, item_changed = _portable_generic_type_arg_with_changed(item)
+        portable_items.append(portable_item)
+        changed = changed or item_changed
+    return (portable_items, True) if changed else (value, False)
+
+
+@_portable_generic_type_arg_with_changed.register(tuple)
+def _(value: tuple) -> tuple[Any, bool]:
+    portable_items, changed = _portable_generic_type_args(value)
+    return (portable_items, True) if changed else (value, False)
+
+
+def _portable_generic_type_args(args: tuple[Any, ...]) -> tuple[tuple[Any, ...], bool]:
+    portable_args = []
+    changed = False
+    for arg in args:
+        portable_arg, arg_changed = _portable_generic_type_arg_with_changed(arg)
+        portable_args.append(portable_arg)
+        changed = changed or arg_changed
+    return tuple(portable_args), changed
+
+
+def _portable_generic_type_arg(value: Any) -> Any:
+    return _portable_generic_type_arg_with_changed(value)[0]
+
+
+@singledispatch
+def _restore_generic_type_arg(value: Any) -> Any:
+    return value
+
+
+@_restore_generic_type_arg.register(_GenericTypeSpec)
+def _(value: _GenericTypeSpec) -> Any:
+    origin, args = value
+    restored_args = tuple(_restore_generic_type_arg(arg) for arg in args)
+    try:
+        return origin[restored_args]
+    except TypeError as exc:
+        if len(restored_args) == 1:
+            try:
+                return origin[restored_args[0]]
+            except TypeError:
+                pass
+        # ``types.UnionType`` is not itself subscriptable; rebuild PEP 604
+        # unions from their members if one of those members needed restoration.
+        union_type = getattr(types, "UnionType", None)
+        if union_type is not None and origin is union_type:
+            result = restored_args[0]
+            for arg in restored_args[1:]:
+                result = result | arg
+            return result
+        raise exc
+
+
+@_restore_generic_type_arg.register(list)
+def _(value: list) -> list:
+    return [_restore_generic_type_arg(item) for item in value]
+
+
+@_restore_generic_type_arg.register(tuple)
+def _(value: tuple) -> tuple:
+    return tuple(_restore_generic_type_arg(item) for item in value)
+
+
+def _new_ccflow_generic_model(origin: Type[PydanticBaseModel], args: Tuple[Any, ...]) -> PydanticBaseModel:
+    """Restore a Pydantic generic specialization without a process-local global."""
+
+    cls = origin[tuple(_restore_generic_type_arg(arg) for arg in args)]
+    return cls.__new__(cls)
+
+
+def reduce_generic_model_instance(model: PydanticBaseModel) -> Optional[tuple[Any, tuple[Any, ...], dict[str, Any]]]:
+    """Return a portable reducer for Pydantic generic specializations."""
+
+    if not _is_pydantic_generic_specialization(type(model)):
+        return None
+
+    metadata = type(model).__pydantic_generic_metadata__
+    args, _ = _portable_generic_type_args(metadata["args"])
+    return (_new_ccflow_generic_model, (metadata["origin"], args), model.__getstate__())
--- a/ccflow/tests/test_base_cloudpickle.py
+++ b/ccflow/tests/test_base_cloudpickle.py
@@ -1,0 +1,321 @@
+import json
+import subprocess
+import sys
+import textwrap
+from pathlib import Path
+
+
+def _run_python(script: str) -> subprocess.CompletedProcess[str]:
+    return subprocess.run([sys.executable, "-c", script], capture_output=True, text=True, timeout=30)
+
+
+def _create_payloads_in_fresh_process(creator_body: str) -> dict[str, dict[str, str]]:
+    # The bug only appears when the receiver has not already materialized the
+    # same Pydantic generic specialization. Create payloads in a subprocess so
+    # the pytest process cannot accidentally warm the receiver's generic cache.
+    creator = _run_python(
+        "import base64\n"
+        "import cloudpickle\n"
+        "import json\n"
+        "import pickle\n"
+        f"{textwrap.dedent(creator_body)}\n"
+        "encoded = {}\n"
+        "for name, (serializer, value) in payloads.items():\n"
+        "    module = pickle if serializer == 'pickle' else cloudpickle\n"
+        "    encoded[name] = {\n"
+        "        'serializer': serializer,\n"
+        "        'payload': base64.b64encode(module.dumps(value, protocol=5)).decode(),\n"
+        "    }\n"
+        "print(json.dumps(encoded, sort_keys=True))\n"
+    )
+    assert creator.returncode == 0, creator.stderr
+    return json.loads(creator.stdout)
+
+
+def _assert_payloads_load_in_fresh_process(creator_body: str, assertions: str, *, extra_loader_setup: str = "") -> None:
+    encoded_payloads = _create_payloads_in_fresh_process(creator_body)
+    # First load each payload in its own fresh process. The original bug was
+    # sensitive to whether a worker had already materialized the same generic
+    # specialization, so a shared loader process can accidentally warm the
+    # receiver and create false positives.
+    for name, spec in encoded_payloads.items():
+        cold_loader = _run_python(
+            "import base64\n"
+            "import cloudpickle\n"
+            "import json\n"
+            "import pickle\n"
+            f"{textwrap.dedent(extra_loader_setup)}\n"
+            f"spec = json.loads({json.dumps(spec)!r})\n"
+            "module = pickle if spec['serializer'] == 'pickle' else cloudpickle\n"
+            "module.loads(base64.b64decode(spec['payload']))\n"
+        )
+        assert cold_loader.returncode == 0, f"{name}: {cold_loader.stderr}"
+
+    # Load in a second fresh process, then run all assertions there. Keeping many
+    # payloads in one process pair keeps the detailed assertions cheap. The
+    # per-payload loop above already verifies cold-receiver loading.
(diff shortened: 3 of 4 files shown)


ptomecek:
Thanks for the very thorough writeup — the problem statement and the cross-process repro/test harness are great. I'd like to propose a much smaller implementation that I believe covers a strict superset of cases.

## Proposal: ~6 lines on the metaclass

Add a `__reduce__` to `_SerializeAsAnyMeta` that reduces any concrete pydantic generic specialization to `origin[args]`:

```python
import operator

class _SerializeAsAnyMeta(ModelMetaclass):
    # ... existing __new__ ...

    def __reduce__(cls):
        md = getattr(cls, "__pydantic_generic_metadata__", None)
        if md and md.get("origin") is not None and md.get("args"):
            return (operator.getitem, (md["origin"], md["args"]))
        return super().__reduce__()
```

The existing `__getstate__` / `__setstate__` on `BaseModel` already handle instance state correctly, so no `__reduce_ex__` override on the instance is needed.

## Why this is sufficient

Pickle saves an instance by first saving its class. For a class object, pickle dispatches on `type(cls)` — the metaclass — and falls through to `__reduce__`. Our reducer returns `(operator.getitem, (origin, args))`, i.e. "reconstruct as `origin[args]` in the receiver." Pickle then **recursively** saves `args`. Each entry in `args` that is itself a generated generic class hits the same metaclass `__reduce__`. That recursion is what removes the need for the singledispatch helpers, the `_PydanticGenericTypeSpec` / `_GenericAliasTypeSpec` wrappers, the `typing` name recovery, and the `types.UnionType` rebuild fallback.

Concretely, the cases the PR description enumerates are all handled by the recursive pickle pass alone:

- `GenericResult[int]` — top-level class reduces to `GenericResult[(int,)]`.
- `GenericResult[ListResult[int]]` — outer reduces to `GenericResult[(ListResult[int],)]`; pickle then reduces `ListResult[int]` via the same hook.
- `GenericResult[list[ListResult[int]]]` / `typing.List[...]` — `list[...]` is a normal pickleable alias; its arg `ListResult[int]` reduces via the hook.
- `GenericResult[Callable[[ListResult[int]], int]]` — `Callable[...]` pickles normally; the embedded `ListResult[int]` reduces via the hook. No need to special-case the parameter list.
- `GenericResult[GenericContext[int] | None]` — `types.UnionType` pickles normally; the embedded `GenericContext[int]` reduces via the hook. No need for the `union_type` rebuild fallback.

## Bonus: it also covers the field-value case the PR explicitly skips

The PR notes that `GenericResult[type](value=ListResult[int])` is intentionally not fixed because the implementation only walks the instance's type args. With a class-level reducer, pickle hits the hook anywhere it encounters such a class in the object graph — including model field values — so that case is fixed for free.

## Verified empirically

I ran a fresh-process reproduction (producer process serializes; cold consumer process only imports origin classes, never materializes the specializations) against the metaclass-only fix for: `simple`, `nested`, `list[...]`, `Callable[[...], int]`, `Optional[...]`. All 5 pass. Happy to share the repro script.

## What I'd suggest keeping from this PR

The implementation can shrink significantly, but the **test file is the most valuable part of the PR** and should stay essentially as-is. The subprocess harness in `test_base_cloudpickle.py` is exactly the right way to guard against regressions, since the bug is invisible in single-process tests. Worth keeping the field-value test too (and we can flip its expectation from "skipped" to "passes").

NeejWeej:
> Thanks for the very thorough writeup — the problem statement and the cross-process repro/test harness are great. I'd like to propose a much smaller implementation that I believe covers a strict superset of cases.
> 
> ## Proposal: ~6 lines on the metaclass
> Add a `__reduce__` to `_SerializeAsAnyMeta` that reduces any concrete pydantic generic specialization to `origin[args]`:
> 
> ```python
> import operator
> 
> class _SerializeAsAnyMeta(ModelMetaclass):
>     # ... existing __new__ ...
> 
>     def __reduce__(cls):
>         md = getattr(cls, "__pydantic_generic_metadata__", None)
>         if md and md.get("origin") is not None and md.get("args"):
>             return (operator.getitem, (md["origin"], md["args"]))
>         return super().__reduce__()
> ```
> 
> The existing `__getstate__` / `__setstate__` on `BaseModel` already handle instance state correctly, so no `__reduce_ex__` override on the instance is needed.
> 
> ## Why this is sufficient
> Pickle saves an instance by first saving its class. For a class object, pickle dispatches on `type(cls)` — the metaclass — and falls through to `__reduce__`. Our reducer returns `(operator.getitem, (origin, args))`, i.e. "reconstruct as `origin[args]` in the receiver." Pickle then **recursively** saves `args`. Each entry in `args` that is itself a generated generic class hits the same metaclass `__reduce__`. That recursion is what removes the need for the singledispatch helpers, the `_PydanticGenericTypeSpec` / `_GenericAliasTypeSpec` wrappers, the `typing` name recovery, and the `types.UnionType` rebuild fallback.
> 
> Concretely, the cases the PR description enumerates are all handled by the recursive pickle pass alone:
> 
> * `GenericResult[int]` — top-level class reduces to `GenericResult[(int,)]`.
> * `GenericResult[ListResult[int]]` — outer reduces to `GenericResult[(ListResult[int],)]`; pickle then reduces `ListResult[int]` via the same hook.
> * `GenericResult[list[ListResult[int]]]` / `typing.List[...]` — `list[...]` is a normal pickleable alias; its arg `ListResult[int]` reduces via the hook.
> * `GenericResult[Callable[[ListResult[int]], int]]` — `Callable[...]` pickles normally; the embedded `ListResult[int]` reduces via the hook. No need to special-case the parameter list.
> * `GenericResult[GenericContext[int] | None]` — `types.UnionType` pickles normally; the embedded `GenericContext[int]` reduces via the hook. No need for the `union_type` rebuild fallback.
> 
> ## Bonus: it also covers the field-value case the PR explicitly skips
> The PR notes that `GenericResult[type](value=ListResult[int])` is intentionally not fixed because the implementation only walks the instance's type args. With a class-level reducer, pickle hits the hook anywhere it encounters such a class in the object graph — including model field values — so that case is fixed for free.
> 
> ## Verified empirically
> I ran a fresh-process reproduction (producer process serializes; cold consumer process only imports origin classes, never materializes the specializations) against the metaclass-only fix for: `simple`, `nested`, `list[...]`, `Callable[[...], int]`, `Optional[...]`. All 5 pass. Happy to share the repro script.
> 
> ## What I'd suggest keeping from this PR
> The implementation can shrink significantly, but the **test file is the most valuable part of the PR** and should stay essentially as-is. The subprocess harness in `test_base_cloudpickle.py` is exactly the right way to guard against regressions, since the bug is invisible in single-process tests. Worth keeping the field-value test too (and we can flip its expectation from "skipped" to "passes").

## TLDR

---

The reconstruction recipe in your suggestion is right: rebuilding as `origin[args]` is exactly what we want. The issue seems to be the hook point. In my repros, `pickle`/`cloudpickle` do not appear to call `__reduce__` on the metaclass when serializing the generated class object itself.

That means a repro can pass for a different reason: if `cloudpickle` takes a by-value path, the fresh consumer does not need the generated `GenericResult[int]` global at all. This lines up with the top-level/function-scope caveat from issue #210.

When I isolate the metaclass-only patch against the ccflow subprocess matrix, the cases from your comment still fail in a cold consumer (`GenericResult[int]`, nested `ListResult[int]`, `Callable[...]`, `Optional[...]`, and the field-value case). The tests inside the new `ccflow/tests/test_base_cloudpickle.py` file also fail. The original branch’s instance-level `BaseModel.__reduce_ex__` path does handle these because that reducer is invoked when pickling the model instance.

So I think the simplification has the right rebuild idea, but not the right hook point. Could you share your repro script? I’d like to check whether it also passes on main / without the patch, or whether there’s another reducer path active that I’m missing.

---

Thanks! This is very helpful. I tried reducing the branch to the metaclass-only implementation and I think I found the discrepancy, but I would like to compare against your repro before drawing a firm conclusion.

The rebuild recipe itself looks right: if `_SerializeAsAnyMeta.__reduce__` is called directly, returning `(operator.getitem, (origin, args))` reconstructs `origin[args]` as intended.

Where I am getting stuck is that, in my isolated repros, `pickle` / `cloudpickle` do not appear to call the metaclass `__reduce__` when serializing the generated class object. Instrumenting both `__reduce__` and `__reduce_ex__` on the metaclass shows zero calls.

Here is a standalone pydantic-shaped version of that check:

```python
import base64
import subprocess
import sys
import tempfile
from pathlib import Path


models_code = """
from typing import Generic, TypeVar

from pydantic import BaseModel
from pydantic._internal._model_construction import ModelMetaclass

T = TypeVar("T")
CALLS = []


class ReproMeta(ModelMetaclass):
    def __reduce__(cls):
        CALLS.append(("__reduce__", cls.__name__))
        md = getattr(cls, "__pydantic_generic_metadata__", None)
        if md and md.get("origin") is not None and md.get("args"):
            import operator
            return (operator.getitem, (md["origin"], md["args"]))
        return super().__reduce__()

    def __reduce_ex__(cls, protocol):
        CALLS.append(("__reduce_ex__", cls.__name__, protocol))
        return super().__reduce_ex__(protocol)


class Box(BaseModel, Generic[T], metaclass=ReproMeta):
    value: T
"""


with tempfile.TemporaryDirectory() as tmp:
    module_path = Path(tmp) / "pydantic_repro_models.py"
    module_path.write_text(models_code)

    producer = f"""
import base64
import cloudpickle
import sys

sys.path.insert(0, {tmp!r})
import pydantic_repro_models as models

# Top-level specialization materialization, matching the risky path from issue #210.
payload = cloudpickle.dumps(models.Box[int](value=5), protocol=5)
print("producer reducer calls:", models.CALLS)
print(base64.b64encode(payload).decode())
"""

    producer_result = subprocess.run(
        [sys.executable, "-c", producer],
        capture_output=True,
        text=True,
        timeout=30,
    )
    print(producer_result.stdout.splitlines()[0])
    encoded_payload = producer_result.stdout.splitlines()[1]

    consumer = f"""
import base64
import cloudpickle
import sys

sys.path.insert(0, {tmp!r})
import pydantic_repro_models as models

print("consumer has Box[int] before:", hasattr(models, "Box[int]"))
obj = cloudpickle.loads(base64.b64decode({encoded_payload!r}))
print(obj)
"""

    consumer_result = subprocess.run(
        [sys.executable, "-c", consumer],
        capture_output=True,
        text=True,
        timeout=30,
    )
    print("consumer returncode:", consumer_result.returncode)
    print("consumer stdout:", consumer_result.stdout.strip())
    print("consumer stderr:", consumer_result.stderr.strip())
```

This prints the important reducer signal:

```text
producer reducer calls: []
```

and the cold consumer fails by generated-name lookup:

```text
consumer has Box[int] before: False
AttributeError: Can't get attribute 'Box[int]' on <module 'pydantic_repro_models' ...>
```

So this is the same pydantic generic shape as the ccflow issue, but the proposed metaclass reducer is still not the hook that `cloudpickle` calls.

The confusing part is that some fresh-process repros can still pass, but I believe that is due to `cloudpickle` choosing a by-value path, not because the metaclass reducer ran. This lines up with the top-level evaluation caveat from issue #210, especially Repro 2 (which failed the same way with the Metaclass __reduce__ approach):

- If `GenericResult[int]` is evaluated at module top level, Pydantic registers the generated specialization as a module global. `cloudpickle` may then serialize by global reference, and a cold consumer can fail because that process-local generated global does not exist.
- If the specialization is only created inside a function, Pydantic may not register it as a module global. `cloudpickle` can then serialize the dynamic class by value, and the fresh-process case can pass.

So I think the issue #210 caveat is the important thing to preserve in any repro comparison: a passing fresh-process case may only show that `cloudpickle` avoided the process-local global-reference path. It does not necessarily show that the metaclass reducer was used.

For ccflow, the metaclass-only patch in `ccflow/base.py` still fails in the cold-consumer path. I checked the same clean cases called out in your comment in the tracked subprocess matrix in `ccflow/tests/test_base_cloudpickle.py`, and also added the field-value case there:

- `GenericResult[int]`
- `GenericResult[ListResult[int]]`
- `GenericResult[list[ListResult[int]]]`
- `GenericResult[Callable[[ListResult[int]], int]]`
- `GenericResult[Optional[ListResult[int]]]`
- `GenericResult[type](value=ListResult[int])`

Those still fail in a cold consumer with errors like:

```text
AttributeError: Can't get attribute 'GenericResult[int]' on ccflow.result.generic
```

and similarly for the nested/callable/optional/field-value cases.

The focused test command I used for the ccflow matrix was:

```bash
python -m pytest ccflow/tests/test_base_cloudpickle.py::test_ccflow_generic_specializations_pickle_across_fresh_processes -q
```

The reducer itself is sanity-checked directly in `ccflow/tests/test_base_serialize.py`, but that only proves the recipe works when called manually; it does not prove `pickle` / `cloudpickle` call it.

So my current read is: the simplification has merit in the reconstruction recipe, but the proposed hook point does not seem to be the mechanism that `pickle` / `cloudpickle` actually uses for these generated class objects. It is possible your repro has another reducer path active, or it is hitting the by-value `cloudpickle` path because of where the specialization is created, or I am mistaken somewhere.

Could you share the exact repro script you used? I want to compare it directly before deciding whether to keep the current instance-level reducer approach or explore a smaller alternative, maybe via a registered class/metaclass reducer path.

ptomecek:
Ok, I see what happened. When testing in `__main__`, pickle/cloudpickle takes a different path that just works, which was independent of this change. 

The problem I have with these changes is that it adds 150-200 lines of complex code to base.py to deal with a very specific corner case. Can you think about putting all this logic in a new module, i.e. `pickling.py` (to mirror `serialization.py`) or something similar. Keep most of the helpers private to that module, expose only what you need publicly, and then try to keep the changes to base.py to <10 lines?

ccflow/base.py
```diff
@@ -165,6 +184,155 @@ def __new__(self, name: str, bases: Tuple[type], namespaces: Dict[str, Any], **k
...
+    args: Tuple[Any, ...]
+
+
+class _GenericAliasTypeSpec(NamedTuple):
+    # Builtin/typing aliases such as ``list[GenericResult[int]]`` are normally
+    # pickleable, but their nested Pydantic generic args may not be. Only wrap
+    # aliases when at least one nested arg had to become portable data.
+    origin: Any
+    args: Tuple[Any, ...]
+    typing_name: Optional[str]
```
ptomecek: **A+B: collapse the two spec types into one and drop `typing_name`.**

`_PydanticGenericTypeSpec` and `_GenericAliasTypeSpec` are nearly identical in shape; the only difference is the `typing_name` field, which exists to round-trip `typing.List` → `typing.List` instead of normalizing to `list`. Pydantic normalizes `typing.List[X]` and `list[X]` to the same core schema during model build, so after a pickle round trip the reconstructed model behaves identically either way — only `type(...).__pydantic_generic_metadata__["args"]` differs in surface form.

If we accept that normalization, both NamedTuples collapse into a single `_GenericTypeSpec(origin, args)` with one restore function (apply `origin[args]` after recursively restoring args, with the existing `types.UnionType` `|`-rebuild fallback). That removes ~30 lines, one `singledispatch` registration, and one place the two paths could drift apart. The only test impact is loosening `typing_list_alias_result`'s assertion from `args == (List[ListResult[int]],)` to an equivalence check (or just `args == (list[ListResult[int]],)`).

ccflow/base.py
```diff
@@ -165,6 +184,155 @@ def __new__(self, name: str, bases: Tuple[type], namespaces: Dict[str, Any], **k
...
+@_restore_generic_type_arg.register(_GenericAliasTypeSpec)
+def _(value: _GenericAliasTypeSpec) -> Any:
+    origin, args, typing_name = value
+    restored_args = tuple(_restore_generic_type_arg(arg) for arg in args)
+    if typing_name is not None:
+        alias = getattr(typing, typing_name)
+        if typing_name == "Optional":
+            non_none_args = tuple(arg for arg in restored_args if arg is not type(None))
+            if len(non_none_args) == 1:
+                return alias[non_none_args[0]]
```
ptomecek: **C: this branch looks like dead code.**

`Optional[X]` evaluates to `Union[X, None]` at definition time, and the resulting `typing._GenericAlias` has `_name = None` (it's a `Union`, not an `Optional`). So `typing_name == "Optional"` is never true at runtime, and the `optional_alias_result` test case passes through the general `origin[restored_args]` path below — which handles `Union[(X, NoneType)]` correctly.

Either remove the branch, or add a test that proves it's reachable. (And if A+B above is taken, this disappears naturally along with the `typing_name` field.)

ccflow/base.py
```diff
@@ -165,6 +184,155 @@ def __new__(self, name: str, bases: Tuple[type], namespaces: Dict[str, Any], **k
...
+        )
+
+    origin = get_origin(value)
+    args = get_args(value)
+    if origin is not None and args:
+        # Example: ``list[ListResult[int]]`` is a normal generic alias, but it
+        # contains a fragile Pydantic specialization. Rebuild the alias from a
+        # portable version of its args only when recursion changed something.
+        portable_args = tuple(_portable_generic_type_arg(arg) for arg in args)
+        if portable_args != args:
```
ptomecek: **D: prefer a `changed` flag over `portable_args != args`.**

Equality on tuples of `typing` constructs has had quirks across Python/pydantic versions (some aliases delegate `__eq__` to origins, some don't; `Annotated`'s equality semantics in particular have moved). Threading a `changed` boolean back from the recursive helper makes the "did anything actually need wrapping?" decision purely structural and removes the dependence on `typing` equality. Cheap to do; mildly clarifying.

ptomecek (commented):
Inline comments below cover the tidy-up opportunities I mentioned in the thread comment. They're independent of the architectural points (move helpers to a `pickling.py` module; the metaclass-`__reduce__` simplification I retracted after NeejWeej's analysis) and would apply on top of any of those. Most leverage is in the first comment — A+B+C in the thread — which collapses the two spec types into one.

ccflow/base.py
```diff
@@ -334,6 +502,21 @@ def __setstate__(self, state):
...
+            # Pydantic's default reducer may serialize runtime generic classes
+            # by names like ``GenericResult[int]``. Those names exist only
+            # after a process has materialized that exact specialization, so
+            # Ray/fresh workers can fail to import them while unpickling.
+            # Serialize a stable constructor instead: generic origin plus
+            # portable generic args. Leave Pydantic instance state to the outer
+            # pickle stream so shared references, cycles, and protocol-5 buffers
+            # keep normal pickle semantics.
+            metadata = type(self).__pydantic_generic_metadata__
+            args = tuple(_portable_generic_type_arg(arg) for arg in metadata.get("args", ()))
```
ptomecek: **E (nit): `metadata.get("args", ())` default is dead defense.**

The `_is_pydantic_generic_specialization` guard above already ensures we only land here when `metadata["origin"] is not None`, which in pydantic v2 implies a non-empty `args` tuple. `metadata["args"]` is sufficient. Same applies on line 237.

ccflow/tests/test_base_serialize.py
```diff
@@ -1,15 +1,39 @@
...
-from ccflow import BaseModel, NDArray
+from ccflow import BaseModel, CallableModelGenericType, GenericContext, GenericResult, NDArray, NullContext
+from ccflow.base import (
+    _GenericAliasTypeSpec,
+    _is_pydantic_generic_specialization,
+    _new_ccflow_generic_model,
+    _portable_generic_type_arg,
+    _PydanticGenericTypeSpec,
+    _restore_generic_type_arg,
+)
```
ptomecek: **Strong test-to-implementation coupling.**

These imports lock the tests to the *internal data shapes* of the implementation (the two NamedTuple types and four helper functions). That means tidy-ups like the A+B collapse on `base.py` become test-breaking changes for non-behavioral reasons.

Consider replacing the spec-shape assertions with a behavioral check: pickle-round-trip the alias and assert the restored alias is equivalent (`==` or by constructing a fresh `GenericResult[alias]` and comparing). That decouples these tests from the spec representation and lets the implementation evolve without churning the test file.

(This also matters for the move to `ccflow/pickling.py` discussed in the thread — these imports would all need re-pointing otherwise.)

ccflow/tests/test_base_cloudpickle.py
```diff
@@ -0,0 +1,321 @@
...
+        assert type(values["local_payload"]).__name__ == "LocalPayload"
+        """,
+        extra_loader_setup=f"""
+        import sys
+        sys.path.insert(0, {str(tmp_path)!r})
+        """,
+    )
+
+
+def test_ccflow_generic_result_cloudpickle_in_ray_worker_without_receiver_materialization():
```
ptomecek: Do we really need this test if we have the tests above? Seems excessive and it's nicer not to have to bring ray into tests unless absolutely necessary.

ccflow/tests/test_base_cloudpickle.py
```diff
@@ -0,0 +1,321 @@
+import json
+import subprocess
+import sys
+import textwrap
+from pathlib import Path
+
+
+def _run_python(script: str) -> subprocess.CompletedProcess[str]:
```
ptomecek: Can we drop the embedded-Python-string subprocess harness in favor of a small run_in_fresh_process helper backed by multiprocessing.get_context("spawn"), so producer/consumer are normal picklable functions rather than strings?

NeejWeej:
> Ok, I see what happened. When testing in `__main__`, pickle/cloudpickle takes a different path that just works, which was independent of this change.
> 
> The problem I have with these changes is that it adds 150-200 lines of complex code to base.py to deal with a very specific corner case. Can you think about putting all this logic in a new module, i.e. `pickling.py` (to mirror `serialization.py`) or something similar. Keep most of the helpers private to that module, expose only what you need publicly, and then try to keep the changes to base.py to <10 lines?

Agreed, done. I moved the generic pickling machinery into `ccflow/pickling.py`; `BaseModel.__reduce_ex__` now just delegates to `reduce_generic_model_instance()` and falls back to Pydantic's reducer when that returns `None`.

ccflow/base.py
NeejWeej: Agreed, done. I collapsed `_PydanticGenericTypeSpec` and `_GenericAliasTypeSpec` into one `_GenericTypeSpec(origin, args)` representation and removed `typing_name`.


This does intentionally allow metadata spelling normalization, e.g. `typing.List[X]` can come back as `list[X]` and `typing.Callable[[X], Y]` can come back as `collections.abc.Callable[[X], Y]`. I updated the tests to treat that as acceptable because the actual model behavior is preserved and we do not need to keep Pydantic's private generic metadata spelling byte-for-byte identical.

ccflow/base.py
NeejWeej: Agreed, done. Removing `typing_name` made this branch go away entirely.

ccflow/base.py
NeejWeej: Agreed, done. The guarded generic paths now use `metadata["args"]`.

ccflow/tests/test_base_serialize.py
NeejWeej: Agreed, done. `test_base_serialize.py` no longer imports the deleted private helper types/functions from `ccflow.base` or asserts exact private spec shapes. I kept a narrow blast-radius check that the reducer only takes over Pydantic generic specializations, and moved the detailed coverage to cold-process behavioral tests.

ccflow/base.py
NeejWeej: Agreed, done. The recursive conversion now returns (value, changed) internally, so the wrapping decision no longer depends on typing equality semantics.

ccflow/tests/test_base_cloudpickle.py
NeejWeej: Agreed, done. I removed the Ray-specific generic pickle test after replacing its important assertion with a spawned-process cold-loader check. The replacement imports the generic origin, verifies `GenericResult[int]` is not already materialized in that worker, then loads the payload and checks the restored value and generic args.

That keeps coverage for the failure mode without making this file depend on Ray startup behavior.

ccflow/tests/test_base_cloudpickle.py
NeejWeej: Tried this and it worked, so done. The cold-process tests now use `multiprocessing.get_context("spawn")` with importable top-level factory/assertion workers instead of `python -c` strings.

I kept the important isolation property: payloads are created in a spawned process, each payload is first loaded in its own fresh spawned process, and the detailed assertions run in a separate spawned process.

ccflow/pickling.py
```diff
@@ -0,0 +1,126 @@
+"""Pickle helpers for ccflow model classes."""
```
ptomecek: Sorry, one last request. Can you embed the pydantic open issue in this module to serve as motivation for why this code currently exists (and so someone knows when it can eventually be deleted)?

NeejWeej:
The low diff coverage here is expected and a little misleading. A large part of this diff is the new cold-
  process pickle test harness. Those helpers intentionally run in spawned child interpreters so we can prove
  deserialization works before the receiver has materialized `GenericResult[int]`.

  Our current coverage setup does not collect coverage from those spawned child processes, so Codecov marks
  those helper lines, and some `ccflow/pickling.py` lines exercised only in those children, as uncovered even
  though the tests are passing.

ccflow/pickling.py
NeejWeej: No worries, added!

ptomecek (approved):

PR #215, opened 2026-05-25
Compatibility fix for Python 3.14 behavior change, Pydantic 2.13

timkpaine:
Closes #127

### DIFF
--- a/ccflow/base.py
+++ b/ccflow/base.py
@@ -11,6 +11,7 @@ import warnings
 from types import GenericAlias, MappingProxyType
 from typing import TYPE_CHECKING, Any, Callable, ClassVar, Dict, Generic, List, Optional, Tuple, Type, TypeVar, Union, get_args, get_origin
 
+import pydantic
 from packaging import version
 
 if TYPE_CHECKING:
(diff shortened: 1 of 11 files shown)


ccflow/base.py
```diff
@@ -106,19 +107,33 @@ def get_registry_dependencies(self, types: Optional[Tuple["ModelType"]] = None)
...
+
+    try:
+        import annotationlib
+    except ImportError:
+        return {}
+
+    annotate = annotationlib.get_annotate_from_class_namespace(namespaces)
+    if annotate is None:
+        return {}
+    return dict(annotationlib.call_annotate_function(annotate, annotationlib.Format.VALUE))
```
ptomecek: Why Format.VALUE instead of Format.FORWARDREF, which I think will handle unresolved forward refs better?

ptomecek (commented):
Why not use the polymorphic_serialization config dict approach from #117 instead of overwriting model_dump and model_dump_json? I think that approach was cleaner - just need to link it to the pydantic version

ptomecek (approved):

PR #219, opened 2026-05-28
Bump version: 0.8.3 → 0.8.4

timkpaine:


ptomecek:
The cache keys will change due to the tokenization work, so this might need to be 0.9.0...

timkpaine:
> The cache keys will change due to the tokenization work, so this might need to be 0.9.0...

I think cache regeneration should be expected on any version change, even patch version. I can add to (forthcoming) documentation that this is the current consideration.

ptomecek (approved):

PR #227, opened 2026-06-04
Add RetryEvaluator

timkpaine:


ptomecek:
The four mechanisms listed for "selecting which models to retry" — `FlowOptionsOverride(model_types=…/models=…)`, per-call `_options`, `meta.options`, and `include_model_types`/`exclude_model_types` — overlap heavily, and the first three already cover essentially every scope (instance, type, per-call, pinned). The only scenario the in-evaluator fields uniquely handle is "I've composed retry into a single global evaluator chain alongside logging/caching and want to disable just the retry component for some model types" — and even there you can solve it by switching between two chains via `FlowOptionsOverride`.

Two things that make me lean toward dropping them:
1. No other evaluator (`LoggingEvaluator`, `MemoryCacheEvaluator`, …) carries per-evaluator selection fields, so this is asymmetric — if we keep it here, the precedent says every future evaluator should grow the same two fields.
2. `exclude_model_types` in particular has no obvious motivating case the override mechanism can't already express.

If you want to keep them, could the docstring spell out the single use case they uniquely solve, so it's clear why retry deserves the extra surface area when nothing else does?

ptomecek:
Minor UX nit on the failure message in `_run_with_retry`. When the loop terminates via the `max_delay` budget check (the `break` on line ~698), the final error reads:

```
Retry attempts exhausted after N attempt(s) for X.
```

But in that path the attempts weren't exhausted — the wait budget was. Could the message distinguish the two cases? Something like:

```
Retry stopped after N attempt(s) for X: max_delay budget exceeded.
```
vs.
```
Retry attempts exhausted after N/N attempt(s) for X.
```

`RetryError.attempts` already carries the count, so the wording is the only thing that needs to change.

ptomecek:
Small doc nit: in the `RetryEvaluator` and `RetryModel` examples in `Key-Features.md`, `retry_exceptions` is shown as strings:

```python
retry_exceptions=["builtins.TimeoutError", "builtins.ConnectionError"],
```

That works (and is the right form for YAML/Hydra configs), but in Python the bare exception class also validates correctly via `PyObjectPath`:

```python
retry_exceptions=[TimeoutError, ConnectionError],
```

which reads much more naturally for a Python user skimming the snippet. Worth showing the bare-class form in the Python example and noting that strings are the form to use in config files.

ptomecek (approved):

PR #231, opened 2026-06-08
Add reporting, retry lifecycle, and dry-run functionality

timkpaine:
Introduce a structured reporting layer that captures evaluation metadata, timing, topology, retries, and failures without consuming result payloads, mirroring the existing evaluator/policy architecture.

- ReportingPolicy shared core with span/run ContextVars for nested, thread/async-local isolation
- ReportEvent model plus NoOp/InMemory/Logging/Composite/UI reporters and a bounded UI polling buffer
- Tracing/metrics/alerts policies and OpenTelemetry tracing/metrics integration (exposed via otel/full/develop/test extras)
- Structural reporting models and a <Vendor><Signal>Reporting{Evaluator,Model} taxonomy with placeholder vendor classes
- Refactor LoggingEvaluator onto LoggingPolicy to share formatting and enable LoggingModel while preserving the existing import path and log output
- Retry lifecycle events now carry run_id and child depth via current_span_depth(); reporter failures are isolated on reporting/retry paths
- DryRunEvaluator with context-local planning guard; synthetic mode is non-transparent so results are not cached under real-run keys; node_key strips the dry-run evaluator layer while preserving non-evaluator options so it matches cache_key() for the logical node
- ReportingStateStore preserves terminal outcomes while allowing retry streams to progress
- Docs: reporting workflow, reporter options, OpenTelemetry install, reserved run/graph phases, extra payload keys, and dry-run synthetic-result warnings
- Tests across utils/evaluators/models covering success/error flows, dry-run override recursion, cache composition, concurrent dry-run reuse, node-key semantics, retry event nesting, reporter failure isolation, and state folding

### DIFF
--- a/ccflow/evaluators/reporting.py
+++ b/ccflow/evaluators/reporting.py
@@ -11,14 +11,9 @@ deduplication. Reporting describes the *evaluation* (model identity, context, ti
 failures) -- it never acts on the result payload, which is what distinguishes it from a publisher.
 
 The class taxonomy is ``<Vendor><Signal>ReportingEvaluator`` where ``Signal`` is one of ``Tracing``,
-``Metrics`` or ``Alerts``. OpenTelemetry tracing/metrics are implemented; the remaining vendor
-backends (Datadog, Opsgenie, JSM, NewRelic, ...) are placeholders that raise ``NotImplementedError``.
+``Metrics`` or ``Alerts``. OpenTelemetry tracing and metrics are implemented.
 """
 
-from collections import deque
-from contextvars import ContextVar
-
-from pydantic import Field
 from typing_extensions import override
 
 from ..callable import EvaluatorBase, ModelEvaluationContext, ResultType
@@ -28,31 +23,18 @@ from ..utils.reporting import (
     MetricsPolicy,
     OpenTelemetryMetricsPolicy,
     OpenTelemetryTracingPolicy,
-    ReportContext,
     ReportingPolicy,
-    ReportPhase,
     TracingPolicy,
-    _new_span_id,
 )
 
 __all__ = [
     "ReportingEvaluator",
-    "DryRunEvaluator",
     "TracingReportingEvaluator",
     "MetricsReportingEvaluator",
     "AlertsReportingEvaluator",
     "OpenTelemetryTracingReportingEvaluator",
     "OpenTelemetryMetricsReportingEvaluator",
     "OpenTelemetryEvaluator",
-    "DatadogTracingReportingEvaluator",
-    "DatadogMetricsReportingEvaluator",
-    "DatadogAlertsReportingEvaluator",
-    "OpsgenieMetricsReportingEvaluator",
-    "OpsgenieAlertsReportingEvaluator",
-    "JSMAlertsReportingEvaluator",
-    "NewRelicTracingReportingEvaluator",
-    "NewRelicMetricsReportingEvaluator",
-    "NewRelicAlertsReportingEvaluator",
 ]
 
 
@@ -81,12 +63,6 @@ def _descriptor(context: ModelEvaluationContext) -> dict:
     }
 
 
-# Re-entrancy guard for DryRunEvaluator, kept context-local (not on the instance) so a single
-# evaluator instance shared across threads / concurrent evaluations cannot leak planning state from
-# one evaluation into another (which would cause the second to run bodies instead of planning).
-_DRY_RUN_PLANNING: ContextVar[bool] = ContextVar("ccflow_dry_run_planning", default=False)
-
-
 class ReportingEvaluator(EvaluatorBase, ReportingPolicy):
     """Evaluator that reports telemetry about the evaluation of the callable models in its scope.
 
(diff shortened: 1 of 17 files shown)


ccflow/evaluators/reporting.py
```diff
@@ -0,0 +1,285 @@
...
+
+    def is_transparent(self, context: ModelEvaluationContext) -> bool:
+        return True
+
+    @override
+    def __call__(self, context: ModelEvaluationContext) -> ResultType:
+        return self._run_with_reporting(context, **_descriptor(context))
+
+
+class DryRunEvaluator(EvaluatorBase, ReportingPolicy):
```
ptomecek: DryRunEvaluator feels a bit out of place here. It's really a planning/preview tool rather than telemetry, but it's sitting in reporting.py and inheriting ReportingPolicy - I don't think anyone hunting for "dry run" would think to look under reporting, and inheriting the policy means it drags reporting config along even when you just want the plan.

It does emit QUEUED/SKIPPED so the dependency on reporting is real, but could we move it to common.py (or its own module) and compose reporting instead of inheriting it? At a minimum I'd like a note explaining why it lives here, since the placement isn't obvious.

ccflow/evaluators/reporting.py
```diff
@@ -0,0 +1,285 @@
...
+class OpenTelemetryMetricsReportingEvaluator(MetricsReportingEvaluator, OpenTelemetryMetricsPolicy):
+    """Metrics evaluator backed by OpenTelemetry counters/histograms (requires ``opentelemetry-api``)."""
+
+
+#: Convenience alias for the most common OpenTelemetry use case (tracing).
+OpenTelemetryEvaluator = OpenTelemetryTracingReportingEvaluator
+
+
+# *****************************************************************************
+# Vendor placeholders (not yet implemented)
```
ptomecek: Can we hold off on shipping these? Right now there are 9 classes that only raise NotImplementedError, and they're all in __all__ - that's more dead names than working ones in this module's public surface, and we don't ship stub classes like this anywhere else in ccflow.

I'd rather add each one when its backend actually exists. If the point is just to stake out the taxonomy, a comment or a private list is a lot lighter than 9 exported classes we'd then have to keep around and eventually deprecate if the plan changes. Config can still reference an unimplemented path via PyObjectPath without a shipped class.

ccflow/utils/reporting.py
```diff
@@ -0,0 +1,907 @@
...
+            span_id=ctx.span_id,
+            parent_span_id=ctx.parent_span_id,
+            run_id=_CURRENT_RUN.get(),
+            depth=ctx.depth,
+            **kwargs,
+        )
+
+    # -- lifecycle hooks (override in subclasses) ----------------------------
+
+    def _on_start(self, ctx: ReportContext, span: "Optional[Span]") -> None:
```
ptomecek, line 497: Since LoggingEvaluator is the default evaluator, this hook runs on basically every model evaluation - and _event() builds a full ReportEvent here even when reporter is None, which is the common case. _emit then just discards it. I measured it at roughly 11us/eval across the three events (START/SUCCESS/END), all wasted when nobody's reporting.

Could we short-circuit when self.reporter is None before constructing the event? The inline logging in LoggingPolicy still needs to fire, but the base emit hooks shouldn't have to build anything if there's no sink.

ptomecek (commented):
Took a pass through this. I really like how the reporting policy/evaluator/model split mirrors the retry trio (RetryPolicy/RetryEvaluator/RetryModel) - it slots in cleanly and reads exactly the way I'd expect. A few things I'd like to sort out before it goes in, left inline.

One non-code note: the description has a "Docs: reporting workflow, reporter options, ..." bullet, but I don't see any docs/ changes in the diff. Did those get dropped, or is the description just ahead of the code? Given how much public API this adds, I think we'll want the docs to land with it.

ccflow/evaluators/reporting.py
timkpaine: Addressed in 4a5b5e8. DryRunEvaluator now lives in its own dry_run module and composes a ReportingPolicy through its reporting field instead of inheriting reporting behavior.

ccflow/evaluators/reporting.py
timkpaine: Addressed in 4a5b5e8. The nine unimplemented public evaluator classes and exports are removed; the possible future backends remain only as a comment until implementations exist.

ccflow/utils/reporting.py
timkpaine: Addressed in 4a5b5e8. Lifecycle hooks now return before constructing ReportEvent instances when no reporter is configured, while LoggingPolicy inline logging still runs. Regression coverage verifies zero event construction on the no-reporter path.

timkpaine:
Addressed review feedback in 4a5b5e8:

- moved dry-run planning to its own evaluator module and composed reporting
- removed unimplemented public backend placeholders, retaining future work as comments
- skipped structured event construction when no reporter is configured
- added reporting API reference documentation and navigation
- removed separator-style comment markers from the changed reporting code

Validation: 75 focused reporting tests pass; Ruff, mdformat, codespell, and diff checks pass.

ptomecek (approved):

PR #259, opened 2026-08-23
Fix cache_key collapse of non-pydantic private/__main__ objects

ptomecek:
Two different objects could produce the same `cache_key`, so the cache could return a result computed for a different input. This fixes that.

### Cause

The generic fallback in `normalize_token` (`ccflow/utils/tokenize.py`) tokenized any object whose type module started with `_` or contained `._` by module and qualname alone, discarding the instance state. This was meant to give a stable token to interpreter internals encountered during behavior hashing.

The condition is too broad. `"__main__".startswith("_")` is `True`, and private submodules such as `pkg._internal` contain `._`. As a result, any non-pydantic object defined in a notebook, REPL, or private module was reduced to its module and qualname, and distinct values produced identical tokens:

```python
class Add:
    def __init__(self, n): self.n = n
    def __call__(self, x): return x + self.n
Add.__module__ = "pkg._private"

compute_data_token(Add(5)) == compute_data_token(Add(7))  # True
```

pydantic models, functions, methods, and partials have dedicated handlers and were unaffected; only the fallback path was involved.

### Change

Decide based on whether the object can be serialized rather than on its module name. If cloudpickle can serialize the object, its state is folded into the token, and the two `Add` instances above tokenize distinctly.

The module name is consulted only when serialization fails, to distinguish two cases:

- Behavior-irrelevant interpreter internals, such as `_abc._abc_data` (surfaced in ABC class closures on Python 3.14 and not picklable), receive a stable name-only token so behavior hashing does not fail.
- Any object in `__main__`, or any other object that cannot be serialized, raises `TypeError` rather than returning a colliding key.

A short allowlist keeps pydantic's compiled validators name-only. They are picklable but carry volatile runtime state that should not enter the token.

Because the decision is based on serializability rather than a fixed list of internal modules, it holds across Python 3.11 through 3.14 without further maintenance.

### Rationale

I instrumented the original branch and ran the full suite to see what reached it. The only objects were the library's own internal helpers, one of which was a stateful frozen dataclass being name-collapsed in the same way (a latent instance of this bug, now also fixed). The `_abc` and pydantic-internal cases are specific to Python 3.14. This is why an allowlist is the wrong approach: it would require ongoing maintenance and would not have covered the library's own modules.

### Tests and validation

Added `TestPrivateModuleNoStateCollapse`, covering:

- picklable instances in `_secret`, `pkg._internal`, and `__main__` tokenize distinctly and deterministically;
- a stateful frozen dataclass in a private module tokenizes distinctly;
- an unpicklable user object in `__main__` raises `TypeError`;
- an unpicklable interpreter-internal object receives a stable name-only token;
- a picklable pydantic-core-style internal is name-only.

The full suite passes on Python 3.11 (1336 passed, 2 skipped), the core `normalize_token` logic was re-checked on Python 3.12, and ruff passes.

### DIFF
--- a/ccflow/utils/tokenize.py
+++ b/ccflow/utils/tokenize.py
@@ -74,6 +74,32 @@ def _with_cycle_check(obj: Any, build: Callable[[], Any]) -> Any:
             _visited.reset(token)
 
 
+# Frameworks whose *private* submodules expose compiled/derived objects that are picklable but carry
+# volatile runtime state (e.g. ``pydantic._internal._model_construction``, ``pydantic_core._pydantic_core``,
+# ``pydantic.plugin._schema_validator``). Key these by module + qualname only, never fold their bytes.
+# Matched by top-level package plus any underscore-prefixed path component, so public modules such as
+# ``pydantic.fields`` are unaffected and unrelated user packages are never captured.
+_FRAMEWORK_INTERNAL_TOP_LEVEL = ("pydantic", "pydantic_core")
+
+# Unpicklable interpreter internals that are behavior-irrelevant and safe to key by name alone: either
+# derived state already captured elsewhere in the hash (``_abc._abc_data``, surfaced in ABC-derived
+# classes' closures on Python 3.14) or primitives with no semantic identity (``_thread`` lock/RLock).
+# Anything else that fails to serialize raises loudly rather than silently collapsing distinct objects
+# (e.g. DB connections, sockets, or user classes in a private ``pkg._internal`` module) onto one key.
+_NAME_ONLY_UNPICKLABLE_MODULES = ("_abc", "_thread")
+
+
+def _is_framework_internal(module: str) -> bool:
+    """Return whether ``module`` is a private submodule of a known framework (see the list above)."""
+    parts = module.split(".")
+    return parts[0] in _FRAMEWORK_INTERNAL_TOP_LEVEL and any(part.startswith("_") for part in parts)
+
+
+def _module_in(module: str, prefixes: tuple[str, ...]) -> bool:
+    """Return whether ``module`` equals or is a submodule of any prefix in ``prefixes``."""
+    return any(module == prefix or module.startswith(prefix + ".") for prefix in prefixes)
+
+
 @singledispatch
 def normalize_token(obj: Any) -> Any:
     """Produce a canonical, deterministically hashable representation of ``obj``.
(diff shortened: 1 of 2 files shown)


ccflow/utils/tokenize.py
```diff
@@ -102,6 +107,11 @@ def _(obj):
     try:
         pickled = cloudpickle.dumps(obj)
     except Exception as exc:
+        # Objects that cannot be serialized. Behavior-irrelevant interpreter internals (e.g. Python 3.14
+        # exposes _abc._abc_data in ABC-derived classes' closures) get a stable name-only token so behavior
+        # hashing doesn't crash; a __main__ object is user code, so it must fail loud instead of colliding.
+        if (obj_module.startswith("_") or "._" in obj_module) and obj_module != "__main__":
+            return ("__internal__", obj_module, type(obj).__qualname__)
```
timkpaine: This treats `__main__` as user code and any other underscore as non-user code, but its reasonable we might have something like `company._models`, etc. We could consider an explicit list of known used interpreter internals, as opposued to the heuristic of `_` prefix.

ptomecek: Good call — replaced the `_`-prefix heuristic with two explicit curated allowlists, and dug into what actually belongs on them.

I probed the common unpicklable objects. The old failure-path heuristic was already inconsistent: `_thread.lock` and `_abc._abc_data` got name-collapsed, but `threading.Event`, `socket.socket`, `sqlite3.Connection`, generators, and weakrefs all fail loud today — purely because their module names don't start with `_`. So real resources mostly already raise; only two things slipped into name-only.

New behavior:

- **Private framework internals** (matched by top-level package `pydantic`/`pydantic_core` + any `_`-prefixed path component) stay name-only. This is picklable-but-volatile compiled state — validators/serializers. Note this version of pydantic puts the validator in `pydantic.plugin._schema_validator`, which my first narrow list missed; the package+private-component match now covers it, `pydantic._internal.*`, and `pydantic_core._pydantic_core` while leaving public modules like `pydantic.fields` alone.
- **Unpicklable interpreter internals** `_abc` (the 3.14 `_abc_data` case) and `_thread` (lock/RLock — primitives with no semantic identity; collapsing them is correct and avoids a regression for any hashed closure that captures a lock) degrade to a stable name-only token.
- **Everything else that can't serialize** — DB connections, sockets, threads, and crucially a user class in `company._models` — now raises `TypeError` instead of silently colliding.

Added tests covering each bucket, including the real `__pydantic_validator__`/`__pydantic_serializer__` objects and a `company._models` case. Full suite green on 3.11.

